PR #34, opened 2015-11-07
base64 encode and decode content when disabling USE_ENCRYPTION

ryan-lane:
If we don't use base64 when USE_ENCRYPTION is set to False, we get control character issues when json.loads is used. We had chosen to not base64 encode the data to make it very obvious in the database that encryption wasn't used, so in this change we're doing a base64, but we're prepending it with UNENCRYPTED_, so that if you look directly at the contents it's clear it's insecure.

### DIFF
--- a/confidant/ciphermanager.py
+++ b/confidant/ciphermanager.py
@@ -1,3 +1,6 @@
+import base64
+import re
+
 from cryptography.fernet import Fernet
 
 from confidant import app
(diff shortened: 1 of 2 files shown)


ryan-lane:
@woodrow for review

ryan-lane:
I also added some dockerfile fixes because the build/test cycle was killing me. These changes make the layers cache more reliably.

confidant/ciphermanager.py
```diff
@@ -22,7 +25,7 @@ def encrypt(self, raw):
             log.warning('Not using encryption in CipherManager.encrypt'
                         ' If you are not running in a development or test'
                         ' environment, this should not be happening!')
-            return raw
+            return 'UNENCRYPTED_{0}'.format(base64.b64encode(raw))
```
woodrow: What would you think about making the unencrypted prefix a constant instead of hardcoding it twice here?

Also, thoughts on something more obvious? `DANGER_NOT_ENCRYPTED_{0}`? :)

ryan-lane:
Since it's only two spots, I'm ok with not having a constant.

woodrow:
+1

PR #40, opened 2015-11-11
Use logging directly and reorder imports for pep8

ryan-lane:
This change switching from using a logging object to using logging directly. This fixes the logging issues in docker. This change also reorders the imports to match pep8 style.


ryan-lane:
@woodrow

woodrow:
+1

PR #43, opened 2015-12-23
Fix index title by correctly referencing index.headline

mkinsella:
The documentation homepage title is currently "Confidant: translation missing: en.index.sub_title" because it is referencing a missing key `index.sub_title`. I changed it to use `index.headline`.

#### Reviewers
@ryan-lane


ryan-lane:
+1 Thanks for the PR!

PR #45, opened 2016-02-24
Add python-virtualenv to apt-get docs.

ab:
You'll need to install virtualenv in order to run virtualenv in the
subsequent lines.


ryan-lane:
Good find. Thanks for the contribution!

PR #47, opened 2016-02-29
Define table region explicitly

ryan-lane:
Seems PynamoDB doesn't use AWS_DEFAULT_REGION, but instead has its own method of accessing tables by region, which is specific to the table creation. Set the region explicitly via AWS_DEFAULT_REGION.


angusfretwell:
Hey, any chance we can get this merged? Our office is piled up with secrets because we have nowhere to put them :cry:

ryan-lane:
I still need to make a release to dockerhub, if you happen to be using the docker image.

angusfretwell:
Yeah, I am using Docker. Could you kindly let me know once you've released it? :smile_cat:

PR #48, opened 2016-02-29
Set GOOGLE_AUTH_EMAIL_SUFFIX to empty string by default

ryan-lane:


ryan-lane:
Actually, this is already the default.

PR #52, opened 2016-03-23
Modular user AuthN, support for SAML

ab:
@ryan-lane This is definitely a work in progress, opening a PR to get feedback on the approach here before I do too much more. I used the 1.1 branch as the base since there was a bunch of refactoring; I hope that was a good thing to do.

Summary of changes:

- Add a dependency on python-saml for SAML support.
- Add an ABC AbstractUserAuthenticator that different user authentication modules can inherit from. These classes must implement `log_in()` and `auth_type()`, and may implement `log_out()` or other custom functionality.
- The auth module in use is configured by `USER_AUTH_MODULE` in settings.
- Convert the existing authentication modules into `NullUserAuthenticator` (for `USE_AUTH=false`) and `GoogleOauthAuthenticator`, which is mostly copy pasted from the existing code.
- Add a new authentication module `SamlAuthenticator`, which implements support for SAML. This also requires some custom routes for the callback handlers, which are all namespaced under `/v1/saml/`.
- Add rudimentary support for `/logout`, which either performs a SAML SingleLogOut flow or just clears the session. The `goodbye.html` is just a stub that I blindly cargo culted from `index.html`.

Testing:
- I haven't written any automated tests yet, and I'm sure this breaks the existing tests, but I haven't yet been able to run them.
- I've manually tested all three auth modules (Null, Google OAuth, and SAML), and they all seem to work!
- I'm hoping to have time to write documentation howtos on getting each of OAuth and SAML integrated with Google as the identity provider. They're both finicky, but pretty straightforward if you know how to do it.


confidant/authnz/__init__.py
```diff
@@ -96,8 +74,8 @@ def check_csrf_token():
     # csrf tokens.
     if g.auth_type == 'kms':
         return True
-    token = request.headers.get('X-XSRF-TOKEN')
-    return safe_str_cmp(token, session.get('XSRF-TOKEN'))
+    token = request.headers.get('X-XSRF-TOKEN', '')
```
ryan-lane: I wonder if we should check to see if token is None and return False here if so.

ab: Ah, good catch.

confidant/authnz/__init__.py
```diff
@@ -205,67 +185,57 @@ def decorated(*args, **kwargs):
...
-class UserUnknownError(Exception):
-    pass
+def require_logout_for_goodbye(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if not app.config.get('USE_AUTH'):
+            return f(*args, **kwargs)
 
+        # ideally we would call check_csrf_token, but I don't think logout CSRF
+        # is a serious concern for this application
```
ryan-lane: Agreed. Not worried about mass logouts here :)

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+from authomatic.adapters import WerkzeugAdapter
+
+# saml auth imports
+from onelogin.saml2.auth import OneLogin_Saml2_Auth
+from confidant.lib import cryptolib
+from confidant.utils.misc import dict_deep_update
+
+from confidant.app import app
+
+from . import errors
```
ryan-lane: When we released we killed off all relative imports. I'm not strongly opposed to them, but would be nice if this could be an explicit import from confidant.authnz

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+            module = GoogleOauthAuthenticator
+        elif module_name == 'saml':
+            module = SamlAuthenticator
+        elif module_name == 'null':
+            module = NullUserAuthenticator
+        else:
+            raise ValueError(
+                'Unknown USER_AUTH_MODULE: {!r}'.format(module_name))
+
+    logging.info('Initializing {} user authenticator'.format(module.auth_type))
```
ryan-lane: Nitpick: {0} rather than {}.

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+        if not root_url:
+            raise ValueError("Must provide SAML_CONFIDANT_URL_ROOT")
+        root_url = root_url.rstrip('/')
+
+        # TODO: also support unspecified?
+        name_id_fmt = 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress'
+
+        # Service Provider section
+        sp_data = {
+            'entityId': root_url + '/v1/saml/metadata',
```
ryan-lane: nitpick: .format or os.path.join

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+                password=app.config.get('SAML_SP_KEY_FILE_PASSWORD'))
+        if app.config['SAML_SP_KEY']:
+            sp_has_key = True
+            sp_data['privateKey'] = app.config['SAML_SP_KEY']
+
+        if app.config['SAML_SP_CERT_FILE']:
+            sp_data['x509cert'] = self._load_x509_for_saml(
+                app.config['SAML_SP_CERT_FILE'])
+        if app.config['SAML_SP_CERT']:
+            sp_data['x509cert'] = app.config['SAML_SP_CERT']
```
ryan-lane: What happens if no key or cert is loaded, either from file or environment? SAML auth fails? Does the class throw an exception on object initialization?

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+            idp_data['singleLogoutService'] = {
+                'url': app.config['SAML_IDP_LOGOUT_URL'],
+                'binding': 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect'
+            }
+
+        if app.config['SAML_IDP_CERT_FILE']:
+            idp_data['x509cert'] = self._load_x509_for_saml(
+                app.config['SAML_IDP_CERT_FILE'])
+        if app.config['SAML_IDP_CERT']:
+            idp_data['x509cert'] = app.config['SAML_IDP_CERT']
```
ryan-lane: Similarly here? Does auth fail? Does it throw an exception?

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+        SAML log-in redirect.
+
+        This method initiates the SAML authentication process by directing the
+        browser to forward along an AuthNRequest to the IdP.
+
+        A separate method handles the post-authentication callback, which will
+        hit /v1/saml/consume, processed by consume_saml_assertion().
+        """
+
+        self_page = request.script_root + request.path
```
ryan-lane: Nitpick: format or os.path.join.

confidant/authnz/userauth.py
```diff
@@ -0,0 +1,661 @@
...
+
+        logging.debug('Processing SAML response')
+
+        try:
+            request_id = session['saml_authn_request_id']
+        except KeyError:
+            logging.warning('No saml_authn_request_id in session')
+            resp = jsonify(errors=['invalid_response'],
+                           message='SAML request failed',
+                           reason=('No AuthNRequest ID from SP found ' +
```
ryan-lane: No need for the + when the strings are within parens, it'll automatically append them.

ryan-lane:
Mostly just style and nitpicks in my comments. I think the general approach is great!

pnathan:
@ryan-lane is this PR planned to be in 1.1?

ryan-lane:
@pnathan yep! going to merge it in whenever @ab is ready

pnathan:
:+1:

confidant/authnz/userauth.py
ab: os.path.join isn't appropriate because this is a URL

confidant/authnz/userauth.py
ab: It's a little tricky because SAML, being endlessly configurable, doesn't _require_ you to sign anything. And python-saml doesn't validate anything at initialization, only raising exceptions at runtime.

If you set the `authnRequestsSigned` setting on python-saml and you then don't provide a key/cert, python-saml will raise a runtime exception when it tries to generate the authN request. 

The approach I took here was to set `authnRequestsSigned` and related options if and only if the `SAML_SP_KEY` setting is provided to confidant. So if you provide a key, all the signing options will be set, and if you don't provide a key, no signing options will be set.

I could be convinced that Confidant's settings should more closely follow the underlying python-saml settings, but I thought it seemed bad to have multiple places where you configure the signing. (e.g. I can't think of any reason you would provide a key and _not_ want signing to be enabled.)

If you don't provide a key and don't have signing enabled, then it's typically the IDP's job to reject requests. So if you try to make a SAML AuthN request to the Google IDP, for example, it will reject unsigned requests.

ab: This is all a bit subtle, so seems like a good area to document not just in a PR. :smile:

confidant/authnz/userauth.py
ab: Similar story to the above SP_CERT stuff. IDP messages and assertions may optionally be signed or encrypted or both.

If you've set the `SAML_SECURITY_MESSAGES_SIGNED` option, then python-saml will at runtime reject IDP messages that are not signed, or it will raise a runtime exception if there is no IDP certificate. I didn't add additional validation that the settings seem self-consistent here because it seemed like second guessing validation that only exists in python-saml at runtime right now.

confidant/authnz/userauth.py
ryan-lane: Makes sense. Thanks for clarifying. I agree with your approach.

confidant/authnz/userauth.py
ryan-lane: Sounds good.

ryan-lane:
Awesome work @ab. Thanks again for this!

confidant/routes/saml.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@app.route('/v1/saml/login', methods=['GET'])
+def generate_saml_login_redirect():
+    """
+    Redirect to the SAML login page. You don't normally need to hit this
+    since any page with @authnz.require_auth will redirect to login.
+    """
+    return flask.redirect(
+        authnz.user_mod.login_redirect_url(return_to='/v1/saml/debug'))
```
pnathan: redirect from login to `/debug`?

ab: Yeah it's a little weird to leave that in there, since I was using it for debugging the IdP integration. But the reason this endpoint exists is for debugging SAML. You never hit this page unless you type this URL explicitly. When tracing the HTTP requests it's nice to have separate start and end pages — a bit confusing to start and end at the same page the way the `@require_auth` wrapped endpoints do.

PR #70, opened 2016-05-25
Fix some bugs with NullUserAuthenticator and reworked login URLs

ab:
There are a few bugs from changes in the 1.1 branch fixed here:
- Define is_expired() method for NullUserAuthenticator objects, which don't inherit from AbstractUserAuthenticator.
- Fix require_csrf_token() decorator so it calls the wrapped function when USE_AUTH is false.
- Fix infinite SAML login loop that comes from dumb code in /v1/saml/consume that has a list of URLs that are known to be infinite loops.

https://github.com/lyft/confidant/commit/ab8bc60cdbc819e4c43b7a9b7f1a18ee0d43b321

### DIFF
--- a/confidant/authnz/__init__.py
+++ b/confidant/authnz/__init__.py
@@ -60,7 +60,7 @@ def require_csrf_token(f):
     def decorated(*args, **kwargs):
         # If we're not using auth, there's no point in checking csrf tokens.
         if not app.config.get('USE_AUTH'):
-            return True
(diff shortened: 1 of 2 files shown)


ab:
OK even taking a quick glance at the code, I'm pretty sure there's more to do here. Feel free to take over or wait for me to get to it.

ab:
I kicked the tires on this a little more and didn't find any other bugs.

Issue #81, opened 2016-06-13
On Debian 8.4: 1.0.1 pip cryptography fails to install when using SSL 1.0.1t-1+deb8u2


russmac:
Essentially what is reported here, Although it only effects cryptography==1.0.1.

https://github.com/pyca/cryptography/issues/2973

Downgrading SSL packages to '1.0.1k-3+deb8u5' resolves the issue.

russmac:
Do you guys feel its safe to bump cryptography version incrementally until the issue stops occurring? 

https://github.com/lyft/confidant/blob/master/requirements.txt#L30

ryan-lane:
Yep. If you'd like you can send a PR in. We'll need you to sign the CLA if so (https://oss.lyft.com/cla). Otherwise I'll try to get this in tomorrow. Sorry about the issues in Debian, and thanks for the report!

PR #82, opened 2016-06-14
Bumping cryptography=1.0.1 to 1.2 to mitigiate https://github.com/lyf…

russmac:
…t/confidant/issues/81

1.2 was the lowest version increment that resolved the issue 1.0.2 failed compilation in the same way as 1.0.1


ryan-lane:
Would you mind using 1.2.3 instead? It has better compatibility: https://cryptography.io/en/latest/changelog/#id7

ryan-lane:
References #81

ryan-lane:
Awesome. Thanks!

PR #83, opened 2016-06-15
Add a user authenticator that uses request headers

andrew-d:
This commit adds a user authenticator that reads authentication from
request headers.  This is useful in the case that you have a load
balancer or other reverse proxy that performs authentication, and simply
being able to make a request to Confidant means that the user is
authenticated.

### DIFF
--- a/confidant/authnz/userauth.py
+++ b/confidant/authnz/userauth.py
@@ -36,6 +36,8 @@ def init_user_auth_class(*args, **kwargs):
             module = GoogleOauthAuthenticator
         elif module_name == 'saml':
             module = SamlAuthenticator
+        elif module_name == 'header':
+            module = HeaderAuthenticator
         elif module_name == 'null':
             module = NullUserAuthenticator
         else:
@@ -268,6 +270,84 @@ class NullUserAuthenticator(object):
     def log_in(self):
         # should never be called
         raise NotImplementedError
+
+
+class HeaderAuthenticator(AbstractUserAuthenticator):
+    """
+    User authenticator that pulls user information from HTTP headers.
+    Note that this assumes we're running behind some form of load-balancer or
+    reverse proxy that performs the authentication, and that simply being able
+    to make requests to this service implies that the user is authenticated.
+    """
+
+    def __init__(self):
+        self.username_header = app.config['HEADER_AUTH_USERNAME_HEADER']
+        self.email_header = app.config['HEADER_AUTH_EMAIL_HEADER']
+        self.first_name_header = app.config.get('HEADER_AUTH_FIRST_NAME_HEADER')
+        self.last_name_header = app.config.get('HEADER_AUTH_LAST_NAME_HEADER')
+
+    @property
+    def auth_type(self):
+        return 'header'
+
+    def assert_headers(self):
+        """Asserts that the current request contains the appropriate headers."""
+        if self.username_header not in request.headers:
+            raise errors.UserUnknownError('No username header in request')
+
+        if self.email_header not in request.headers:
+            raise errors.UserUnknownError('No email header in request')
+
+    def current_user(self):
+        self.assert_headers()
+
+        info = {
+            'email': request.headers[self.email_header],
+
+            # TODO: should we use a string like "unknown", fall back to the
+            # email/username, ...?
+            'first_name': '',
+            'last_name': '',
+        }
+
+        if self.first_name_header and self.first_name_header in request.headers:
+            info['first_name'] = request.headers[self.first_name_header]
+
+        if self.last_name_header and self.last_name_header in request.headers:
+            info['last_name'] = request.headers[self.last_name_header]
+
+        return info
+
+    def current_email(self):
+        return self.current_user()['email'].lower()
+
+    def current_first_name(self):
+        return self.current_user()['first_name']
+
+    def current_last_name(self):
+        return self.current_user()['last_name']
+
+    def is_authenticated(self):
+        """Any user that is able to make requests is authenticated"""
+        self.assert_headers()
+        return True
+
+    def is_expired(self):
+        """Sessions are not managed here and do not expire"""
+        return False
+
+    def check_authorization(self):
+        """Header users are always authorized"""
+        self.assert_headers()
+        return True
+
+    def log_in(self):
+        self.assert_headers()
+
+        # Does nothing, since simply being able to reach this endpoint is 'logging in'.
+        resp = self.redirect_to_index()
+        self.set_csrf_token(resp)
+        return resp
 
 
 class GoogleOauthAuthenticator(AbstractUserAuthenticator):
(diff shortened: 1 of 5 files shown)


andrew-d:
Note: this is still somewhat of a WIP - I need to think a bit more about what semantics of this should be.  That being said, I'm putting my initial work up here for any comments or suggestions 😄

ryan-lane:
Hey, this is looking good so far. Thanks for the PR! I'll do a quick initial review tomorrow. Before we're able to merge this it's necessary for you to sign our CLA (https://oss.lyft.com/cla).

Also, looks like the tests are currently failing due to pep8 violations. I know I'm ridiculous, but I keep the line length set to 80.

andrew-d:
CLA signed!  I'll fix the pep8 failures tomorrow :+1:

confidant/authnz/userauth.py
```diff
@@ -270,6 +272,85 @@ def log_in(self):
...
+
+    def current_first_name(self):
+        return self.current_user()['first_name']
+
+    def current_last_name(self):
+        return self.current_user()['last_name']
+
+    def is_authenticated(self):
+        """Any user that is able to make requests is authenticated"""
+        self.assert_headers()
```
ryan-lane: The assertion is because we want to throw exceptions if the user isn't authenticated? I'm guessing this is because it's a somewhat severe error if the user isn't authenticated?

confidant/authnz/userauth.py
```diff
@@ -270,6 +272,85 @@ def log_in(self):
...
+        return True
+
+    def log_in(self):
+        self.assert_headers()
+
+        # Does nothing, since simply being able to reach this endpoint is
+        # 'logging in'.
+        resp = self.redirect_to_index()
+        self.set_csrf_token(resp)
```
ryan-lane: If users don't need to log in, it's really important that the CSRF token is set whenever the user first accesses confidant. Otherwise, if they go in through some other entrypoint and don't hit the login url, they won't have a CSRF token and any POST/PUT action will fail.

ryan-lane: It's possible that this function is being called when the session is opened, but it would be good to double check that.

tests/unit/confidant/authnz/authnz_test.py
```diff
@@ -209,3 +210,57 @@ def test_redirect_to_logout_if_no_auth(self):
...
+
+    def tearDown(self):
+        app.config['HEADER_AUTH_USERNAME_HEADER'] = self.username_header
+        app.config['HEADER_AUTH_EMAIL_HEADER'] = self.email_header
+        app.config['HEADER_AUTH_FIRST_NAME_HEADER'] = self.first_name_header
+        app.config['HEADER_AUTH_LAST_NAME_HEADER'] = self.last_name_header
+
+        authnz.user_mod = self.user_mod
+
+    def test_will_extract_from_request(self):
```
ryan-lane: This is technically an integration test. Since it doesn't require dynamo or kms access we can leave it in until we get the integration tests working properly in travis.

tests/unit/confidant/authnz/authnz_test.py
```diff
@@ -209,3 +210,57 @@ def test_redirect_to_logout_if_no_auth(self):
...
+
+            # Both headers given: success
+            with patch('confidant.authnz.userauth.request') as request_mock:
+                request_mock.headers = {
+                    app.config['HEADER_AUTH_USERNAME_HEADER']: 'unittestuser',
+                    app.config['HEADER_AUTH_EMAIL_HEADER']: 'unittestuser@example.com',
+                }
+                self.assertEqual(authnz.get_logged_in_user(), 'unittestuser@example.com')
+
+    def test_will_log_in(self):
```
ryan-lane: Also an integration test. Will leave in until we can run the integration test suite in travis.

confidant/authnz/userauth.py
```diff
@@ -270,6 +272,85 @@ def log_in(self):
...
+            raise errors.UserUnknownError('No email header in request')
+
+    def current_user(self):
+        self.assert_headers()
+
+        info = {
+            'email': request.headers[self.email_header],
+
+            # TODO: should we use a string like "unknown", fall back to the
+            # email/username, ...?
```
ryan-lane: I think it's fine to leave these as empty strings, if they aren't supported as headers.

confidant/authnz/userauth.py
ryan-lane: Yeah... log_in isn't called when the session is opened, so you'll need to modify your auth flow in some way where you ensure the csrf token is set and returned.

ryan-lane:
Hey @andrew-d, we're preparing for the 1.1 release and was hoping to get this change in. I think the only thing that needs to get updated is the CSRF token stuff. Did you get a chance to follow-up on this?

andrew-d:
Hi there, and sorry for the wait!  I was out on vacation until Tuesday this week, and have been slowly catching up on things that have accumulated while I was offline.  I'll try to get this in a workable state tomorrow 👍

ryan-lane:
Awesome. Thanks!

ryan-lane:
Hey @doy, thanks for the commit! For us to accept your commit, we'll need to you sign our CLA: https://oss.lyft.com/cla

doy:
Okay, I signed the CLA. I'll write up some tests for this tomorrow, but does this seem like a reasonable approach for the CSRF issues?

ryan-lane:
I think it'll solve the issue, yeah. I'm trying to consider if this will adversely affect SAML or Google Auth. I'll take a closer look soon :)

doy:
Turns out I can't actually add tests for the request flow of arbitrary user auth modules, since it's determined at load time: https://github.com/lyft/confidant/blob/1.1/confidant/authnz/__init__.py#L21. Is the code here sufficient, or is there a better way to test this?

ryan-lane:
It is possible to patch the user_mod on the module with an instance of the module you want to test; see: https://github.com/lyft/confidant/blob/1.1/tests/integration/confidant/authnz/authnz_test.py#L34-L40

That said, we could add a factory function to load the module and cache it as a global so that it's a singleton, which would make it easier than this to test.

doy:
How's this?

ryan-lane:
Looks great and is passing tests. Let me test this a bit with google auth to make sure it's working outside of null/header auth.

doy:
After some more testing, it looks like this doesn't quite work with all endpoints - I'll see if I can fix.

doy:
(In case it wasn't clear, that latest commit should address the issues that came up in my testing.)

doy:
Any updates here?

ryan-lane:
Sorry. Let me fire up a local dev environment with this change and then I'll merge it in :)

ryan-lane:
Everything is working as expected! Merging in. Thanks again for the PR!

PR #105, opened 2017-01-06
Move scripts into module and make admin scripts available via pip ins…

ryan-lane:
…tall


confidant/scripts/manage.py
```diff
@@ -0,0 +1,24 @@
...
+manager.add_command("revoke_all_kms_auth_grants", RevokeGrants)
+
+# Generate encrypted blob from a file
+manager.add_command("generate_secrets_bootstrap", GenerateSecretsBootstrap)
+
+# Show the YAML formatted secrets_bootstrap in a decrypted form
+manager.add_command("decrypt_secrets_bootstrap", DecryptSecretsBootstrap)
+
+if __name__ == "__main__":
+    manager.run()
```
jayaddison: Why `manager.run()` here and `manager.main()` from the root `manage.py`?

Issue #102, opened 2016-10-19
When redirecting user to loggedout also clear the cookies


ryan-lane:
We've run into situations where the CSRF tokens aren't always fully purged for some reason and users end up with two tokens, which causes them failures. When users are logged out, the JS should additionally clear all cookies associated with the domain, as an extra reliability measure.

ryan-lane:
Looks like the cases we're seeing this it's due to conflicting XSRF cookies in a domain, so the proper fix is to use a cookie name that's unique to confidant.

ryan-lane:
Fixed in 1.3.0

PR #106, opened 2017-01-09
Make xsrf cookie name configurable

ryan-lane:
This change makes it possible to modify the xsrf cookie name used by the web frontend. This can be used if the cookie name conflicts with another application in the same domain.


ryan-lane:
Fixes #102

PR #111, opened 2017-01-24
Add a simple webhook

egeland:
Add the ability to notify of credential/service changes via a simple webhook.

Basically a copy and modify of the graphite event notification.

### DIFF
--- a/confidant/webhook.py
+++ b/confidant/webhook.py
@@ -5,11 +5,11 @@ import logging
 from confidant.app import app
 
 
-def send_event(event_type, services, credentials):
+def send_event(event_type, services, credential_ids):
     try:
         webhook_url = app.config.get('WEBHOOK_URL')
         if not webhook_url:
-            logging.warning('Failed to find a WEBHOOK_URL in config')
+            logging.debug('Failed to find a WEBHOOK_URL in config')
             return
         username = app.config.get('WEBHOOK_USERNAME')
         password = app.config.get('WEBHOOK_PASSWORD')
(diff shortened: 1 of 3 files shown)


ryan-lane:
Hey @egeland, thanks for the contribution! I'm going to start reviewing, but to accept the PR we'll need you to sign our CLA: https://oss.lyft.com/cla

Let me know when you sign it and I'll verify :)

confidant/webhook.py
```diff
@@ -0,0 +1,35 @@
...
+import logging
+
+from confidant.app import app
+
+
+def send_event(event_type, services, credentials):
+    try:
+        webhook_url = app.config.get('WEBHOOK_URL')
+        if not webhook_url:
+            logging.warning('Failed to find a WEBHOOK_URL in config')
```
ryan-lane: On second thought, it's likely that most folks won't have webhooks setup. Can you make this a debug, rather than a warning?

confidant/webhook.py
```diff
@@ -0,0 +1,35 @@
+import requests
+import json
+import logging
+
+from confidant.app import app
+
+
+def send_event(event_type, services, credentials):
```
ryan-lane: Can you change `credentials` to `credential_ids` to clarify that folks should send ids and never credentials into this :).

Also, can you add a function doc string talking about the args and the expected input? We really want to make sure no one ever accidentally calls this with actual credentials, and only ever a list of IDs.

egeland:
I have signed the CLA :)

ryan-lane:
Awesome. Merging in. Thanks for the contribution!

ryan-lane:
Released as 1.4.0.

PR #112, opened 2017-02-01
Toggle credential conflict check

egeland:
We have a use case where we need to be able to save credentials that have the same key names, and use them in the one service.

Since the credential itself acts like a kind of namespace, this looks safe to do, and as we're making the default match the current behaviour, only people who want this feature will get it.

To elaborate on the use case, we use the Service to map into a kubernetes namespace, and the Credentials inside this service with names like `kube-cluster-<clustername>` to contain the cluster details to connect to. Those key/value pairs will be the same (i.e `username`/`password`), so we need to turn off the conflict check.

### DIFF
--- a/confidant/routes/v1.py
+++ b/confidant/routes/v1.py
@@ -261,7 +261,7 @@ def map_service_credentials(id):
             data.get('credentials', []),
             data.get('blind_credentials', []),
         )
-        if conflicts:
+        if conflicts and not app.config['IGNORE_CONFLICTS']:
             ret = {
                 'error': 'Conflicting key pairs in mapped service.',
                 'conflicts': conflicts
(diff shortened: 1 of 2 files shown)


egeland:
Hmm.. Travis failed, but seems to be because of `npm`, not because of me (I built the docker image locally, no problems)...?

ryan-lane:
Thanks for the contribution! I'm taking a look. The travis failure was due to an npm timeout, I've triggered a new build.

ryan-lane:
I have a feeling @doy will be interested in this. I think the change looks good, but I want to do some testing with this tomorrow to be sure. We unfortunately don't have tests wrapped around this and I've been meaning to refactor all this logic out of the routes for ages so I'd like to be careful.

ryan-lane:
Looks good. Merging in.

PR #114, opened 2017-02-07
Security fix: bump python-saml version to address vulnerability

egeland:
According to the python-saml repo:

> ### Warning
>
> Update python-saml to 2.2.0, this version includes a security patch that contains extra validations that will prevent signature wrapping attacks. [CVE-2016-1000252](https://github.com/distributedweaknessfiling/DWF-Database-Artifacts/blob/master/DWF/2016/1000252/CVE-2016-1000252.json)
>
> python-saml < v2.2.0 is vulnerable and allows signature wrapping!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -45,7 +45,7 @@ pyflakes==1.5.0           # via flake8
 pynamodb==1.5.3
 pyopenssl==16.2.0
 python-dateutil==2.6.0    # via botocore, delorean
-python-saml==2.1.6
+python-saml==2.2.1
 pytz==2016.10             # via babel, delorean, tzlocal
 PyYAML==3.11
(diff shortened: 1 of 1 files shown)


ryan-lane:
@egeland the requirements are kind of a pain to update. You need to update requirements.in, then run pip-compile (from piptools) to update the requirements.txt. I'll work up a make target for this in the future, but for now let me update this.

egeland:
I'd never used pip-compile before, so learning new things today :)

ryan-lane:
We're using in somewhat of a weird way where we have a container run that does some extra stuff too, so I'm not sure anyone but us can easily update the reqs at this point. I'll need to work on making that open sourced.

PR #123, opened 2017-02-28
change to a lighter docker image by using alpine as the base

egeland:
Switching from ubuntu to alpine makes the image about half the size. :)


egeland:
Some re-jigging to make fewer/thinner layers, too

ryan-lane:
Sorry for not responding earlier. We're not comfortable with using alpine for a lot of reasons (mainly security and compatibility). I think we can probably make the current image much smaller by splitting the frontend generation into another image, rather than switching over to alpine.

PR #128, opened 2017-03-29
Switch to ujson where possible

ryan-lane:


confidant/keymanager.py
```diff
@@ -11,6 +10,7 @@
 import confidant.services
 from confidant.app import app
 from confidant.utils import stats
+from confidant.utils import json as json
```
danielhochman: no reason to as json

danielhochman:
+1

nit: remove all the needless uses of `as json`

confidant/utils/json.py
```diff
@@ -0,0 +1,22 @@
+import logging
+import json
```
danielhochman: kind of weary of having the filename be the same as the json module.

i don't think there's actually a conflict since it's not at top-level, but worth considering a change

danielhochman:
+1

PR #131, opened 2017-03-31
Add manage script to create dynamo tables

Stype:
In some cases (like running tests) we don't want to automatically create tables, but instead create them intentionally at a later time with a manage script.

### DIFF
--- a/confidant/scripts/bootstrap.py
+++ b/confidant/scripts/bootstrap.py
@@ -12,6 +12,7 @@ from flask.ext.script import Option
 from confidant import settings
 from confidant.app import app
 from confidant.lib import cryptolib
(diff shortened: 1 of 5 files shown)


confidant/scripts/bootstrap.py
```diff
@@ -61,3 +62,11 @@ def run(self, _out):
         else:
             with open(os.path.join(_out), 'w') as f:
                 f.write(data)
+
+
+class CreateDynamoTables(Command):
```
ryan-lane: I think this should be in its own util script.

ryan-lane: Bootstrap is for encrypting/decrypting the secrets in confidant's config file.

PR #132, opened 2017-03-31
Remove import causing jailed tests to fail

Stype:
Flake8 says the import is unused, and the import is causing kms to be called from testss


confidant/settings.py
```diff
@@ -13,6 +13,31 @@ class SettingsError(Exception):
...
+        Lazy setup things that we want to get if we have kms
+        """
+        if name not in self.secret_names:
+            self.secret_names.append(name)
+            self.secret_defaults[name] = default
+
+    def registered(self, name):
+        return name in self.secret_names
+
+    def get_for_realz(self, name):
```
ryan-lane: Heh. can we call this lazy_load? :)

tests/unit/confidant/authnz/authnz_test.py
```diff
@@ -3,6 +3,11 @@
 from mock import Mock
 from werkzeug.exceptions import Unauthorized
 
+# Prevent call to KMS when app is imported
+from confidant import settings
+settings.encrypted_settings.secret_string = {}
```
ryan-lane: This is ok for now, but it's likely better to patch these directly in the used functions, so there's not unintended side effects across tests.

PR #133, opened 2017-04-04
Consolidate credential and blind credential (part 1)

ryan-lane:
This change takes a first step towards consolidating credential and blind credential into a single resource type, differentiated by a `blind` attribute. This is to reduce a lot of the duplication in the code and reduce the number of dynamo queries required unnecessarily.

As a side effect, this change also changes the schema for normal credentials for credential_pairs and data_keys, to match the model used by blind credentials. This allows us to encrypt credential_pairs using KMS keys from multiple regions, adding multi-region support.


ryan-lane:
blind credential support is being dropped in a future version, so no need to combine them

PR #138, opened 2017-04-06
Upgrade gunicorn, and other piptools transitive deps

ryan-lane:


ryan-lane:
Fixes https://github.com/lyft/confidant/pull/138

PR #142, opened 2017-04-11
Fix setting config settings from non-bootstrapped configs

Stype:


ryan-lane:
+1 tested this and it fixes the issue

Issue #143, opened 2017-04-11
Bootstrap settings set without SECRETS_BOOTSTRAP broken in master


ryan-lane:
Settings like a SESSION_SECRET are broken in master unless SECRETS_BOOTSTRAP is used. Reported by MaXx in IRC channel.

ryan-lane:
Fixed by https://github.com/lyft/confidant/pull/142

ryan-lane:
Will close when a regression test is added.

PR #144, opened 2017-04-11
Add tests for encrypted settings

Stype:


Stype:
Completes work for https://github.com/lyft/confidant/issues/143

tests/unit/confidant/encrypted_settings_test.py
```diff
@@ -0,0 +1,48 @@
+import unittest
+from mock import patch
+from mock import Mock
+
+# Prevent call to KMS when app is imported
+from confidant import settings
+settings.encrypted_settings.secret_string = {}
+settings.encrypted_settings.decrypted_secrets = {'SESSION_SECRET': 'TEST_KEY'}
```
ryan-lane: Can you do this as part of a setUp/tearDown sequence? We have some other tests to use as examples. I worry that this mocking will affect other tests, depending on the run order.

tests/unit/confidant/encrypted_settings_test.py
```diff
@@ -0,0 +1,48 @@
...
+        assert enc_set.get_secret('Foo') == 'DecryptedBar'
+
+    def test_get_registered_default(self):
+        enc_set = EncryptedSettings(None)
+        enc_set.register('Foo', 'Bar')
+        enc_set.register('Bar', 'Baz')
+        enc_set.decrypted_secrets = {'Foo': 'DecryptedFoo'}
+        assert enc_set.get_secret('Bar') == 'Baz'
+
+    @patch('confidant.encrypted_settings.cryptolib.decrypt_datakey', return_value='1cVUbJT58SbMt4Wk4xmEZoNhZGdWO_vg1IJiXwc6HGs=')
```
ryan-lane: pep8 line length == 80 in confidant. I know, I'm absurd.

ryan-lane:
+1 after setup/teardown stuff

tests/unit/confidant/encrypted_settings_test.py
```diff
@@ -0,0 +1,48 @@
...
+# Prevent call to KMS when app is imported
+from confidant import settings
+settings.encrypted_settings.secret_string = {}
+settings.encrypted_settings.decrypted_secrets = {'SESSION_SECRET': 'TEST_KEY'}
+
+from confidant.app import app
+from confidant.encrypted_settings import EncryptedSettings
+
+
+class EncprytedSettingsTest(unittest.TestCase):
```
ryan-lane: EncryptedSettingsTest

tests/unit/confidant/encrypted_settings_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def test_app_config(self):
+        # This is really an integration test
+        assert app.config.get('SESSION_SECRET') == 'TEST_KEY'
```
ryan-lane: We're using nose, not pytest, so you should use self.assertEqual(a, b)

tests/unit/confidant/encrypted_settings_test.py
Stype: Thinking more on this, not really. The problem is that as soon as confidant.app is imported, it will try to load the encrypted blob from the environment. I'm going to try a followup to make that not happen.

PR #153, opened 2017-10-10
Added check to prevent creating a service that already exists

vivianho:
This checks that if the user is creating a new service, they do not override an existing service.


ryan-lane:
Looks like the travis jobs are failing...

PR #155, opened 2017-10-11
Scale up dynamodb table

adilhafeez:
This change will scale up the table to at least 2x of the load we've seen in the past week (Friday evening specifically),

`INFO:ddb table: confidant-production-iad not scaled up for read, found violation at: 2017-09-30 01:37:00+00:00, consumed capacity: 919, provisioned capacity: 1500`


ryan-lane:
Wrong spot for this.

PR #158, opened 2017-11-02
 Fix issue with angular trimming whitespace

### LINKED ISSUE #156: Confidant strips newlines when submitting credentials through webpage
We're using Confidant to store, among other things, our certificates. When making requests between our machines we noticed that they were failing due to a SSL certificate verification error. Through some investigation, I found that the md5sum of the certificate is different after it's been sent through Confidant. 

It seems like this is caused by the angular.js frontend, which I believe is stripping off newlines. I tried pasting in a value with a newline at the end, confirmed that the blank line was there after pasting, and found it to be missing when editing the credential again after saving.

I then tried to edit the value by adding newlines and noticed that the textarea class is still marked as "ng-pristine". After adding alphabetical characters, you can see that "ng-pristine" is removed and "ng-dirty" is added to the textarea class. I've pasted images below showing the three states, where the textarea has not been edited, a newline has been added but the area is still marked as pristine, and then with text added and the area marked as dirty. 

For reference, we're running Confidant version 1.1.13. 

image

image

image

ryan-lane:
Seems by default for text areas that angularjs trims whitespace. This sets ng-trim to false for credential values. Fixes #156.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # 1
 
+## 1.10.1
+
+* Fix an issue in the angularjs frontend where credential values were having whitespace trimmed.
+
 ## 1.10.0
 
 * Upgrade gevent and greenlet for CVE-2016-5180 and gevent/gevent#477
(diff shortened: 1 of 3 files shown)


lizneu:
+1

PR #160, opened 2017-12-01
Upgrade pynamo to 2.2.0 with migration and maintenance mode

ryan-lane:
This change upgrades pynamodb to 2.2.0, with a migration script for UnicodeSetAttribute. It also adds a maintenance mode that can be used to disable writes via the API, to be able to run a migration uninterrupted.


CHANGELOG.md
```diff
@@ -1,6 +1,26 @@
 # Changelog
 
-# 1
+# 2
+
+## 2.0.0
+
+WARNING: If you upgrade to this version, any new writes to blind credentials
+will be in a format that is only compatible in 1.11.0 forward. If you've
+upgraded and need to downgrade, you should downgrade to 1.11.0.
```
vivianho: Maybe add explicit instructions for how to upgrade? e.g. "If your blind credentials table is not empty, follow these steps"

confidant/scripts/migrate.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+app.logger.addHandler(logging.StreamHandler(sys.stdout))
+app.logger.setLevel(logging.INFO)
+
+
+class MigrateSetAttribute(Command):
+
+    def run(self):
+        app.logger.info('Migrating UnicodeSetAttribute in BlindCredential')
```
vivianho: Do you need to enforce that the service is in maintenance mode before doing this?

ryan-lane: Nah. I think it's OK if we tell folks that it's best if they have it in maintenance mode, but I don't think we need to enforce it.

CHANGELOG.md
ryan-lane: Yeah, I'll add some more info here and will add an "Upgrading" section to the docs.

ryan-lane: Done

ryan-lane:
Rebase looks good.

PR #165, opened 2018-03-10
Add NonNullUnicodeSetAttribute class to deal with None

betamoo:


confidant/models/blind_credential.py
```diff
@@ -16,6 +16,24 @@
 from confidant.models.connection_cls import DDBConnection
 
 
+class NonNullUnicodeSetAttribute(UnicodeSetAttribute):
```
vivianho: Could we extract this into its own module and have the `models.service` and `models.blind_credential` modules import this? So that we don't copy/paste code.

betamoo: sounds right!

vivianho:
@ryan-lane can you review? Should we bump this to version 3.0.1?

confidant/models/blind_credential.py
```diff
@@ -14,6 +13,9 @@
 from confidant.app import app
 from confidant.models.session_cls import DDBSession
 from confidant.models.connection_cls import DDBConnection
+from confidant.models.non_null_unicode_set_attribiute import (
```
vivianho: @betamoo typo here? `non_null_unicode_set_attribiute`

PR #171, opened 2018-04-03
No need for sudo in Dockerfile

asottile:


vivianho:
Can you bump the version to 4.0.1? I'll make a new release.

asottile:
@vivianho this doesn't impact the contents of the python package so no release is necessary here.

PR #176, opened 2018-07-04
Minor Dockerfile updates

wyardley:
* Update `MAINTAINER` to `LABEL maintainer=`
* Bump node version to 8 LTS
* install ca-certificates so curl call doesn't fail because of cert validation
* add `--fail` option to curl so it exits non-0 if something does fail
* add `DEBIAN_FRONTEND=noninteractive` and `--no-install-recommends` to other
  `apt-get` call

I also adjusted some of the formatting to be a little more consistent, and added spaces in the `CMD` array (I think this looks nicer, but could take it out).

I didn't add it here, but might also be interesting to add some of the labels from http://label-schema.org/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
 FROM ubuntu:trusty
-MAINTAINER Ryan Lane <rlane@lyft.com>
+LABEL maintainer="rlane@lyft.com"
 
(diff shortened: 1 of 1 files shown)


ryan-lane:
Looks good. Thanks for the PR!

PR #182, opened 2019-01-23
Case-insensitive sort for credential and service lists

ryan-lane:
Current sorting for credentials isn't case insensitive, which sorts lowercase together and uppercase together, which makes it awkward to find credentials when adding them to a service.


ikonst:
I'm sorry.

ryan-lane:
LOL. Not your fault. We should always ensure master is in a buildable
state. Our internal salt code for this is correct, so I just needed to
reference it to see how to fix things.

PR #184, opened 2019-02-20
Do not use mutable arguments in pynamo models

ryan-lane:
To avoid potential future issues with data corruption, if usage of the models changes, avoid use of mutable args in pynamo models.


ryan-lane:
meow @woodrow 
meow @vivianho 
meow @jmphilli

Issue #186, opened 2019-04-12
More Webhook support


edify42:
The current design only allows for 1 webhook endpoint to be configured (with basic authentication supported).

I'd like to extend this by allowing 1 extra environment variable to be injected `WEBHOOK_ARRAY_FILE` which points to a JSON file with data that looks like below:

```
[
  {
    "WEBHOOK_URL": "https://webhook_dev.example.com"
  },
  {
    "WEBHOOK_URL": "https://webhook_test.example.com",
    "WEBHOOK_USERNAME": "myhookuser",
    "WEBHOOK_PASSWORD": "mylongandsupersecurehookpassword"
  }
]
```

This file can be mounted onto the docker container and the path passed in via `WEBHOOK_ARRAY_FILE`. The existing functionality will not be broken and users can now add multiple webhook endpoints when secrets are updated.

edify42:
@ryan-lane is this something we could do?

I saw the issue @egeland raised with Webhooks regarding security

ryan-lane:
Hey! Let me take a look at the current solution and the security concerns raised. My first gut reaction is that this feels ok, though.

ryan-lane:
Oh, hah. I raised the concerns, right. Let me re-evaluate my thoughts there.

egeland:
Given that the information sent by the webhook would (still) not include any secrets, adding the possibility to send more is a nice feature-add (and something I've thought about as well 😅 ).

edify42:
[Created a PR to tie into this issue](https://github.com/lyft/confidant/pull/187)

Some use-cases for extra webhooks leaving Confidant:

- Messaging notifications upon secrets being changed in Confidant
- CI triggers to update secrets in apps
- Additional environment support (dev and test) for any Webhook integration above

egeland:
Any movement on this, @ryan-lane ? 😉

ryan-lane:
Hey! Sorry, I've been on sabbatical. I'll be back at the end of the month
but may take a quick look before then.

ryan-lane:
@egeland oh, it looks like the PR comments I left a while back haven't been addressed yet.

egeland:
Suspect @edify42 doesn't use Confidant anymore?

edify42:
Indeed @egeland I don't! - Happy for you to take over this PR if you're keen :)

PR #187, opened 2019-04-15
Added support for multiple webhooks declared in a JSON file

edify42:
As mentioned in[this issue](https://github.com/lyft/confidant/issues/186) I want to be able to send multiple webhooks from Confidant when a secret (service or credential) is updated.

I went with a JSON configuration file which can be mounted onto the running container and the `WEBHOOK_ARRAY_FILE` environment variable storing the location of the file which can be retrieved.

The logic is purposefully kept separate from the existing Webhook functionality.

The only other thing possibly worth justifying was the try/except block - I wanted other Webhook list items not to fail if an earlier listed webhook endpoint returned a failed response (e.g. username+password authentication failure or a general 50x backend error).

### DIFF
--- a/confidant/settings.py
+++ b/confidant/settings.py
@@ -379,6 +379,22 @@ WEBHOOK_PASSWORD = encrypted_settings.register(
     'WEBHOOK_PASSWORD',
     str_env('WEBHOOK_PASSWORD')
 )
+# Webhook Array File configuration
+# Example:
+# cat webhook_file.json
+# [
+#   {
+#     "WEBHOOK_URL": "https://webhook_dev.example.com"
+#   },
+#   {
+#     "WEBHOOK_URL": "https://webhook_test.example.com",
+#     "WEBHOOK_USERNAME": "myhookuser",
+#     "WEBHOOK_PASSWORD": "mylongandsupersecurehookpassword"
+#   }
+# ]
+# Extends the above standard webhook configuration to allow for more
+# integrations
+WEBHOOK_ARRAY_FILE = str_env('WEBHOOK_ARRAY_FILE')
 
 # Ignore conflicts of credential names in a service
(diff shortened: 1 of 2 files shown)


confidant/settings.py
```diff
@@ -379,6 +379,22 @@ def str_env(var_name, default=''):
...
+# Example:
+# cat webhook_file.json
+# [
+#   {
+#     "WEBHOOK_URL": "https://webhook_dev.example.com"
+#   },
+#   {
+#     "WEBHOOK_URL": "https://webhook_test.example.com",
+#     "WEBHOOK_USERNAME": "myhookuser",
+#     "WEBHOOK_PASSWORD": "mylongandsupersecurehookpassword"
```
ryan-lane: passwords should really be part of the encrypted settings, otherwise we're storing secrets in the clear somewhere.

ryan-lane:
It would be good to find a way to store the passwords in the encrypted_settings, somehow. Also, please add tests :)

confidant/settings.py
edify42: Hey @ryan-lane thanks for your feedback.

For my clarity, I have a few linked questions:

1. Is it still OK to have the `WEBHOOK_ARRAY_FILE` store the value for `WEBHOOK_PASSWORD` in plain text? I'd like to rely on the container management engine to place the file onto the file system accessible by the confidant container.
2. If (1) is OK, I'm feeling there will be changes to the `settings.py` and `encrypted_settings.py` files to loop through the file and then possibly store the encrypted passwords in a dict.

ryan-lane: The encrypted_settings code is there to load a set of values from a pre-encrypted SECRETS_BOOTSTRAP environment variable. It's there to avoid needing to load any plain text passwords into confidant. I'd prefer to not have confidant load any secrets from plaintext.

PR #188, opened 2019-06-05
Support for SAML roles (readonly / write access for credentials)

jrosco:
Hi, The idea behind this PR is to allow setting SAML roles in the credential metadata which would allow a user with correct SAML role settings read-only or write access to certain credentials. 

A list of features.

- Setup an Admin role to have access to all credentials  (This is set up in the env config file SAML_ADMIN_ROLE)
- Able to disable/enable SAML role permissions (This is set up in the env config file SAML_USE_ROLE)
- Show the users role on the Web GUI
- Able to view role via the API (/v1/user/role)
- Add roles to the credentials metadata (These are set up in the env config file SAML_ROLE_RW_NAME and SAML_ROLE_R_NAME )
- Add the SAML Role name (This is set up in the env config file SAML_MAPPING_ROLE)
- Added unittest

We currently have this setup in our production environment and it's working with no known issues


ryan-lane:
Hey, thanks for the PR! Can you sign the CLA? I'll start taking a look through the PR :)

confidant/routes/v1.py
```diff
@@ -50,14 +49,24 @@ def get_user_info():
         response = jsonify({'email': None})
     return response
 
+@app.route('/v1/user/role', methods=['GET', 'POST'])
+@authnz.require_auth
+def get_role_info():
```
ryan-lane: Any reason not to include this info in the `get_user_info()` endpoint?

confidant/routes/v1.py
```diff
@@ -374,6 +383,10 @@ def get_credential(id):
             'Item with id {0} does not exist.'.format(id)
         )
         return jsonify({}), 404
+
+    if not authnz.require_saml_role(cred.metadata)['read_only']:
```
ryan-lane: I think it's maybe better to have this as a decorator that can be applied to endpoints, rather than being checked in-function.

For example:

```
@authnz.require_role('read_only')
```

confidant/routes/v1.py
```diff
@@ -780,6 +794,10 @@ def update_credential(id):
     credential_pairs = cipher.encrypt(update['credential_pairs'])
     update['metadata'] = data.get('metadata', _cred.metadata)
     update['documentation'] = data.get('documentation', _cred.documentation)
+
+    if not authnz.require_saml_role(_cred.metadata)['read_write']:
+        return jsonify({'error': 'Credential Metadata found Role does not have write permissions'}), 403
```
ryan-lane: Same as above. Would be good to make a decorator for this.

ryan-lane: I think you also need to apply this to `create_credential()`, and `map_service_credentials()`

confidant/authnz/__init__.py
```diff
@@ -238,6 +245,52 @@ def decorated(*args, **kwargs):
     return decorated
 
 
+def require_saml_role(credential):
```
ryan-lane: Would be nice for this to be `require_role`, rather than `require_saml_role`, so other modules can implement it, if necessary.

confidant/routes/v1.py
jrosco: Yeah, I think that would be a better place for this endpoint

confidant/routes/v1.py
jrosco: I will work on this

confidant/authnz/__init__.py
jrosco: Agreed

confidant/routes/v1.py
jrosco: Looking at the get_user_info() it uses the route '/v1/user/email' so doesn't make sense to put role here. Might be better making an endpoint like '/v1/user/userinfo' and adding email and role details here.

ryan-lane: I think it would be ok to rename the endpoint to `/v1/user/userinfo`. No reason it needs to be `/v1/user/email`. I'm pretty sure it's just used by the frontend to display the user's email address, so it should be an easy change.

jrosco:
I need to write some unittest for the `require_role` function

jrosco:
@ryan-lane Are you ok with me closing this PR and reopening a new one since it looks like some of the commits have an Author failing the CLA check.

ryan-lane:
Yep. If all of the code will be yours or if it's associated with someone in
your company and and as such are allowed to grant a copyright license on
their behalf.

jrosco:
will create a new pull request.

PR #190, opened 2019-08-29
Convert confidant to python3

asottile:


tests/unit/confidant/keymanager_test.py
```diff
@@ -169,7 +169,7 @@ def test_decrypt_token(self, kms_mock):
                 1,
                 'service',
                 'confidant-unittest',
-                'ZW5jcnlwdGVk'
+                b'ZW5jcnlwdGVk'
```
vivianho: Just for my reference, was this always passed in / converted to a binary string in the actual code (i.e. here: https://github.com/lyft/confidant/blob/a652892b3664d25216d1b396205bd103eced1360/confidant/authnz/__init__.py#L176) and you don't need to make the same change here?

asottile: it probably wasn't binary before, I'll have to check once I can get it to start and run

you're probably correct that the test change here is incorrect and it _should_ be text and not bytes (and python2 was just like "ah whatever" since it's ascii-only)

when that's the case I'll switch this to do the proper encoding dance for the inputs here :+1: 

I still haven't been able to get this working in our infrastructure due to the wacky javascript build though :upside_down_face:

PR #191, opened 2019-09-09
Switch to kmsauth library for kmsauth support

ryan-lane:
We split kmsauth out to a library a while ago, but hadn't switched confidant over to using it. This change uses kmsauth, rather than having the code in two spots.


confidant/keymanager.py
```diff
@@ -29,40 +25,17 @@
...
-            KeyId='alias/{0}'.format(key_alias)
-        )
-    return KEY_METADATA[key_alias]['KeyMetadata']['Arn']
-
-
 def get_key_id(key_alias):
     if key_alias not in KEY_METADATA:
         KEY_METADATA[key_alias] = auth_kms_client.describe_key(
-            KeyId='alias/{0}'.format(key_alias)
+            KeyId='{0}'.format(key_alias)
```
vivianho: nit: I don't think there's a need to use formatted string here

confidant/keymanager.py
```diff
@@ -77,7 +50,7 @@ def create_datakey(encryption_context):
     stats.incr('at_rest_action', 2)
     return cryptolib.create_datakey(
         encryption_context,
-        'alias/{0}'.format(app.config.get('KMS_MASTER_KEY')),
+        '{0}'.format(app.config.get('KMS_MASTER_KEY')),
```
vivianho: nit: I don't think there's a need to use formatted string here

PR #199, opened 2019-09-23
Feature/role support

jrosco:
The idea behind this PR is to allow setting roles (SAML for now) in the credential metadata which would allow a user with correct role settings read-only or write access to certain credentials.

**A list of features.**

- Setup an Admin role to have access to all credentials (This is set up in the env config file ADMIN_ROLE)
- Able to disable/enable role permissions (This is set up in the env config file USE_ROLE)
- Show the users role on the Web GUI (top right-hand corner)
- Add roles to the credentials metadata (These are set up in the env config file ROLE_RW_NAME and ROLE_R_NAME )
- Add the Role name attribute (This is set up in the env config file ROLE_MAPPING_NAME)
- Combine user email and role into one single API endpoint `/v1/user/info`
- Able to view role via the API (/v1/user/info)

**How to use:**
The roles are set up and mapped in the credentials metadata (shown below)

Screenshot at Sep 24 12-51-55

role_ro = read only 
role_rw = read write

So in the above example, the **user** has read-only access and the **engineer** and **engineer-manager** have read-write access

_N/B There is an Admin role which allows full access to all credentials. This can be set with the **ADMIN_ROLE** environment variable._

**To Do:**

- Add python unit test
- Check if we can add the `require_role` decorator to resources/service
- Support multiple groups for a single user
- Test/Check if this can work with other Auth methods .e.g Goggle, etc


confidant/authnz/__init__.py
```diff
@@ -237,6 +247,51 @@ def decorated(*args, **kwargs):
...
 
+def require_role(role):
+    """
+    Get the users role in credential metadata
+    """
+    def decorated(f):
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            cred_id = kwargs['id']
+            cred = Credential.get(cred_id)
```
ryan-lane: You should defer getting the credential until necessary, otherwise it adds load on the system, even when USE_ROLES is disabled, for example

confidant/routes/v1.py
```diff
@@ -366,6 +370,7 @@ def get_credential_list():
 
 @app.route('/v1/credentials/<id>', methods=['GET'])
 @authnz.require_auth
+@authnz.require_role(role='read_only')
```
ryan-lane: Since you need to check against a resource, it may be better to have this as a function call inside of the function, rather than as a decorator, as you'll be able to pass-in the credential model instance, rather than doing a double fetch.

ryan-lane: Also, if you do it as a function call, you could make it generic and only pass in metadata, which would make the function work with anything that has metadata (credentials, services, blind credentials, etc).

confidant/authnz/__init__.py
ryan-lane: Or, as mentioned in another comment, change this to a function (not a decorator) which takes metadata as an argument, which will let you avoid doing a get on the credential.

confidant/authnz/userauth.py
```diff
@@ -113,11 +113,12 @@ def set_expiration(self):
                 max_expiration = now + datetime.timedelta(seconds=max_lifetime)
                 session['max_expiration'] = max_expiration
 
-    def set_current_user(self, email, first_name=None, last_name=None):
+    def set_current_user(self, **kwargs):
```
ryan-lane: Prefer adding role=None to the kwargs, than using ``**kwargs``. This change as currently made makes all arguments required.

confidant/authnz/__init__.py
```diff
@@ -237,6 +247,51 @@ def decorated(*args, **kwargs):
...
+                    user_role = user_mod.current_role()
+                    role_name_rw = app.config['ROLE_RW_NAME'].lower()
+                    role_name_r = app.config['ROLE_RO_NAME'].lower()
+                    groups_rw = cred.metadata.get(role_name_rw)
+                    groups_r = cred.metadata.get(role_name_r)
+
+                    if user_role is not None:
+                        if user_role == app.config['ADMIN_ROLE'].lower():
+                            return make_response(f(*args, **kwargs))
+                        if role_name_rw in cred.metadata:
```
ryan-lane: nit: elif

confidant/authnz/__init__.py
```diff
@@ -237,6 +247,51 @@ def decorated(*args, **kwargs):
...
+                            groups_rw = groups_rw.replace(' ', '').split(',')
+                            if user_role in groups_rw:
+                                return make_response(f(*args, **kwargs))
+                        if role_name_r in cred.metadata:
```
ryan-lane: nit: elif

confidant/authnz/__init__.py
```diff
@@ -237,6 +247,51 @@ def decorated(*args, **kwargs):
...
+            cred_id = kwargs['id']
+            cred = Credential.get(cred_id)
+
+            if app.config['USE_ROLES']:
+                try:
+                    user_role = user_mod.current_role()
+                    role_name_rw = app.config['ROLE_RW_NAME'].lower()
+                    role_name_r = app.config['ROLE_RO_NAME'].lower()
+                    groups_rw = cred.metadata.get(role_name_rw)
+                    groups_r = cred.metadata.get(role_name_r)
```
ryan-lane: Should each of these be `.lower()` as well? Pretty sure we don't force these to lowercase in the model.

confidant/authnz/userauth.py
```diff
@@ -131,6 +132,12 @@ def current_first_name(self):
     def current_last_name(self):
         return self.current_user()['last_name']
 
+    def current_role(self):
+        if type(self.current_user()['role']) is list:
```
ryan-lane: Doing a type check here seems odd. Shouldn't this be: `if not self.current_user()['role']:`

confidant/authnz/userauth.py
jrosco: fixed

confidant/authnz/__init__.py
jrosco: added elif

confidant/authnz/__init__.py
jrosco: added elif

confidant/authnz/userauth.py
jrosco: fixed

confidant/authnz/__init__.py
```diff
@@ -237,6 +247,58 @@ def decorated(*args, **kwargs):
...
+
+                    if groups_r is not None:
+                        groups_r = groups_r.lower()
+                        groups_r = groups_r.replace(' ', '').split(',')
+
+                    if groups_rw is not None:
+                        groups_rw = groups_rw.lower()
+                        groups_rw = groups_rw.replace(' ', '').split(',')
+
```
jrosco: added lower() function and split and removed white space functions

confidant/authnz/__init__.py
jrosco: fixed below

confidant/routes/v1.py
jrosco: do you mean a function call within the `get_credential()`  and `update_credential()` functions ? 

example:
```
authnz.require_role(Credential.metadata)
```

ryan-lane: yep!

confidant/authnz/__init__.py
```diff
@@ -280,7 +280,8 @@ def wrapper(*args, **kwargs):
                                 or role_name_r in cred.metadata:
                             if groups_rw and user_role in groups_rw:
                                 return make_response(f(*args, **kwargs))
-                            elif groups_r and user_role in groups_r:
+                            elif groups_r and user_role in groups_r \
```
ryan-lane: Use parens, rather than backslash

ryan-lane:
@jrosco In the past few releases, we've introduced a rather thorough ACL/RBAC framework that you can use to implement this in a fine-grained way that suits your needs. See: https://lyft.github.io/confidant/acls.html

The implementation we added is actually more fine-grained and should give you a lot more control :)

PR #206, opened 2019-11-19
Log users and the credentials they obtain

skiptomyliu:
Log users and the credentials that they obtain.  Useful for auditing and alerting.


confidant/routes/v1.py
```diff
@@ -392,6 +392,9 @@ def get_credential(id):
     cipher = CipherManager(data_key, cipher_version)
     _credential_pairs = cipher.decrypt(cred.credential_pairs)
     _credential_pairs = json.loads(_credential_pairs)
+
+    if authnz.user_is_user_type('user'):
+        logging.info(f"{authnz.get_logged_in_user()} fetching credential {id}")
```
ryan-lane: confidant is still python2 compat, so ideally use .format(), rather than f strings

skiptomyliu: Thanks Ryan, good point.

confidant/routes/v1.py
```diff
@@ -392,6 +392,9 @@ def get_credential(id):
     cipher = CipherManager(data_key, cipher_version)
     _credential_pairs = cipher.decrypt(cred.credential_pairs)
     _credential_pairs = json.loads(_credential_pairs)
+
+    if authnz.user_is_user_type('user'):
```
SecPrez: Do we want this to be users only? If an attacker has access to a service and is pulling credentials we would want to know that as well. I would rather over log.

ryan-lane: @SecPrez I think we should be smart about how we log that particular action, or make it configurable or pluggable in some way. It has the potential to have a very large volume, with little usefulness.

It may be good to log the first use of a kmsauth token (it's not cached), or when a token is being used from a new IP address. The latter would require more effort, as you'd need to store a list of IPs along with the tokens.

For the most part, it would be unexpected for a kmsauth token to be reused across IPs, as they're meant to be generated per-client. Tokens are also expected to be cached for some amount of time, as it's slightly expensive to generate and validate them.

SecPrez: I would rather have user only sooner if the service requests are harder to log.

PR #210, opened 2019-12-05
Add function to hook into external ACL function

skiptomyliu:
Add an external hook that allows us to integrate with an external ACL module.


confidant/routes/v1.py
```diff
@@ -371,6 +372,14 @@ def get_credential_list():
 @app.route('/v1/credentials/<id>', methods=['GET'])
 @authnz.require_auth
 def get_credential(id):
+    if not acl_module_check(id):
```
ryan-lane: I think we need to pass in more info than just the ID. I think maybe `acl_module_check('get_credential', 'view', id)`

^^ that way we can make specific checks based on function, the permission within the function, and the resource it's being called against.

confidant/routes/v1.py
```diff
@@ -102,13 +106,21 @@ def get_service(id):
     if authnz.user_is_user_type('service'):
         if not authnz.user_is_service(id):
             logging.warning('Authz failed for service {0}.'.format(id))
-            msg = 'Authenticated user is not authorized.'
+            msg = 'Service is not authorized.'
             return jsonify({'error': msg}), 401
     else:
+        logged_in_user = authnz.get_logged_in_user()
+        if not acl_module_check('get_service',
+                                metadata_only=metadata_only,
```
ryan-lane: I think we should check for allowed permissions here. So for instance, we'd pass in a list maybe? `['view', 'metadata']`

skiptomyliu: Just updated along with adding the `acl_module_check` to additional endpoints. 

Lemme know if this is what ya had in mind.  
```
metadata_only = request.args.get('metadata_only', default=False, type=bool)
...
acl_actions = ['metadata']
if not metadata_only:
    acl_actions.append('get')
```
In the current logic, users default will be able to view full credentials unless `metadata_only` explicitly set to `True`

confidant/routes/v1.py
```diff
@@ -479,6 +533,16 @@ def get_credential_list():
 @app.route('/v1/credentials/<id>', methods=['GET'])
 @authnz.require_auth
 def get_credential(id):
+    if not acl_module_check('get_credential',
+                            actions=['get'],
+                            resource=id):
```
ryan-lane: We should followup in another PR to also support metadata here

confidant/settings.py
```diff
@@ -491,3 +492,21 @@ def get(name, default=None):
     if encrypted_settings.registered(name):
         return encrypted_settings.get_secret(name)
     return globals().get(name, default)
+
+
+def load_module(module_path):
```
ryan-lane: Can we move this to utils?

skiptomyliu: moved to `util.misc`

PR #211, opened 2019-12-06
Use revert endpoint for reverting service and add metadata only fetch

ryan-lane:
This change is aimed at returning less credential data, when that data is not necessary. Specifically, it's aiming to remove the return of full credential data when services are fetched through the interface, since those interfaces don't need the credential data.

Along with this change is a new method of doing reverts. I'm starting with service, because it's simpler, but future changes will also make this the method for credentials and blind_credentials, which will let us do reverts without needing to see values.

### DIFF
--- a/confidant/public/modules/history/controllers/BlindCredentialHistoryCtrl.js
+++ b/confidant/public/modules/history/controllers/BlindCredentialHistoryCtrl.js
@@ -69,6 +69,14 @@
                 }
             });
 
+            $scope.getCredentialByID = function(id) {
+                return $filter('filter')($scope.$parent.credentialList, {'id': id})[0];
+            };
+
+            $scope.getBlindCredentialByID = function(id) {
+                return $filter('filter')($scope.$parent.blindCredentialList, {'id': id})[0];
+            };
+
             $scope.revertToDiffRevision = function() {
                 var deferred = $q.defer();
                 if (angular.equals($scope.diffBlindCredential.name, $scope.currentBlindCredential.name) &&
--- a/confidant/routes/v1.py
+++ b/confidant/routes/v1.py
@@ -104,6 +104,7 @@ def get_service(id):
     Get service metadata and all credentials for this service. This endpoint
     allows basic authentication.
     '''
+    metadata_only = request.args.get('metadata_only', default=False, type=bool)
     if authnz.user_is_user_type('service'):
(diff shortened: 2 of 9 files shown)


ryan-lane:
I also found a bug in displaying conflicts in the history revert view, which I fixed by adding getCredentialByID and getBlindCredentialByID

confidant/public/modules/history/controllers/BlindCredentialHistoryCtrl.js
```diff
@@ -69,6 +69,14 @@
                 }
             });
 
+            $scope.getCredentialByID = function(id) {
+                return $filter('filter')($scope.$parent.credentialList, {'id': id})[0];
```
skiptomyliu: apologies, by JS is rusty.  We can always expect an entry to be returned by this?  Meaning, `[0]` is safe to return?

confidant/routes/v1.py
```diff
@@ -324,18 +366,117 @@ def map_service_credentials(id):
...
+            revert_service.credentials,
+            revert_service.blind_credentials,
+        )
+        if conflicts:
+            ret = {
+                'error': 'Conflicting key pairs in mapped service.',
+                'conflicts': conflicts
+            }
+            return jsonify(ret), 400
+    # Try to save to the archive
```
skiptomyliu: Curious -- how come we save to archive here?   Is it a sort of backup?

confidant/public/modules/history/controllers/BlindCredentialHistoryCtrl.js
ryan-lane: `$scope.$parent.credentialList` is a pre-generated list. If there's no credentials in that list, then nothing would ever call `getCredentialByID`. If `getCredentialByID`, then something must be in the list.

confidant/routes/v1.py
ryan-lane: Confidant's dynamo usage is a bit weird. We use a single table for all record types, and we store history and current records in the same table. For performance, we have two types of record per model. So here, we have `service` and `service-archive` type records. The `service` record is the one that's queried, when `get_service` is called. When history is looked up, it's the archive records.

If we were to split the records into separate tables, we would use hash+key, which would let us drop the double-save. It's something I've considered, if we were to get rid of blind credentials (which we can do with proper RBAC).

PR #212, opened 2019-12-06
Cache iam role list using a background gevent thread

ryan-lane:
In AWS accounts with a large number of IAM roles, the `get_iam_roles_list` endpoint is extremely slow, and can timeout. Also, we tend to call this frequently, and it's difficult to cache in the frontend.

This change caches the IAM role list in-process, and refreshes it on a refresh schedule, in the background.

### DIFF
--- a/confidant/services/iamrolemanager.py
+++ b/confidant/services/iamrolemanager.py
@@ -1,0 +1,50 @@
+import gevent
+import logging
+import random
+
+import confidant.clients
+from confidant import settings
+
+ROLES = []
+
+
+def refresh_cache():
+    global ROLES
+    refresh_rate = settings.BACKGROUND_CACHE_IAM_ROLE_REFRESH_RATE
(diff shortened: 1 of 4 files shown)


confidant/services/iamrolemanager.py
```diff
@@ -0,0 +1,50 @@
...
+from confidant import settings
+
+ROLES = []
+
+
+def refresh_cache():
+    global ROLES
+    refresh_rate = settings.BACKGROUND_CACHE_IAM_ROLE_REFRESH_RATE
+    if settings.BACKGROUND_CACHE_IAM_ROLE_REFRESH_RATE < 60:
+        refresh_rate = 60
```
skiptomyliu: nit: Should we make the 60 a configurable constant in settings?

skiptomyliu: Eh, nm.  I think okay.  I see there's a  +- 20 secs which makes things a bit more complicated.

PR #226, opened 2019-12-13
Support metadata_only for credential

skiptomyliu:
Similar to `get_service`, provide the `GET` option for omitting the credential pair.


confidant/routes/v1.py
```diff
@@ -479,6 +479,7 @@ def get_credential_list():
 @app.route('/v1/credentials/<id>', methods=['GET'])
 @authnz.require_auth
 def get_credential(id):
+    metadata_only = request.args.get('metadata_only', default=False, type=bool)
```
ryan-lane: For this we don't want the client to determine this, but instead determine what level of permissions the user has, and just give them what they're allowed to get

ryan-lane: We want the client to determine it for get_service, because the UI doesn't need the sensitive data, but service clients do.

PR #229, opened 2019-12-14
Use github actions, rather than travis

ryan-lane:
This change switches the CI from travis to github actions. It also ensures the frontend tests run (linting), fixes lint errors that existed, and switches python linting over to pre-commit (and adds precommit). This change will also auto-generate and publish the docs on master merges.


ryan-lane:
A nice plus of this change is that tests run ~3-4m faster

PR #231, opened 2019-12-18
Non-sensitive diff support for credential history

ryan-lane:
This change modifies the behavior and data in the history view for credentials, to remove sensitive information from that view. A new diff_credential endpoint has been added, which will take two revisions for a credential ID, and returns a non-sensitive diff of the two credentials.

The history view for the credential shows a -/+ view of the two credentials, side-by-side. For credential_pairs (the sensitive part), it shows what's been added and removed, by credential_pair key. In the case a credential pair has been modified, it'll show as being removed and added (similar to a git diff).

### DIFF
--- a/confidant/models/credential.py
+++ b/confidant/models/credential.py
@@ -65,6 +65,55 @@ class Credential(Model):
             return False
         return True
 
+    def diff(self, other_cred):
+        if self.revision == other_cred.revision:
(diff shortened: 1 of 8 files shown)


ryan-lane:
Screenshot 2019-12-17 15 52 44

PR #234, opened 2019-12-20
Refactor resources view to match history view for resource toggling

ryan-lane:
This change refactors the resources view to be consistent with the history view for resources. Rather than showing all resources of one type in the same list, there's a toggle view between credentials, blind credentials, and services. I also pulled the create toggle into the same row as the toggles, and away from the filter, which was confusing for folks in the past.

This change also adds links to a resource's history from the details view. It also styles the action callout buttons (create, edit, save, revert) to make them more apparent.

### DIFF
--- a/confidant/public/modules/app.js
+++ b/confidant/public/modules/app.js
@@ -22,8 +22,12 @@
      * Main controller
      */
     .controller('ConfidantMainCtrl', [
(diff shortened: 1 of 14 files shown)


ryan-lane:
Screen Shot 2019-12-20 at 4 13 18 PM
Screen Shot 2019-12-20 at 4 13 38 PM

PR #235, opened 2019-12-21
Refactor RBAC acl_module_check calls

skiptomyliu:
Simplify and standardize the RBAC calls to `acl_module_check(resource_type, actions, resource_id, **kwargs)`.

This makes integrating with the module much more predictable

Examples:
```
acl_module_check('credential', ['metadata', 'get'], '12345')
acl_module_check('service', ['get'], 'abc-service-role')
acl_module_check('service', ['list'])
acl_module_check('service', ['update'], 'abc-service-role', {'credential_ids': ['123']})
```

### DIFF
--- a/confidant/routes/v1.py
+++ b/confidant/routes/v1.py
@@ -80,7 +80,7 @@ def get_client_config():
 @authnz.require_auth
 def get_service_list():
     if not acl_module_check(resource_type='service',
-                            actions=['list']):
+                            action='list'):
         msg = "{} does not have access to list services".format(
             authnz.get_logged_in_user()
         )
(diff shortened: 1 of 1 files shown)


ryan-lane:
I know I originally asked for the actions to be a list, but I think it's probably better as just a string. It's not super clear what the list means from the end-user point of view (does it mean AND, or OR?). I think it's clearer to have it be explicit, then if we want to do AND or OR, we can do it at the calling location.

skiptomyliu:
Sounds good.  I'l refactor.

ryan-lane:
Actually, let's close this in favor of: https://github.com/lyft/confidant/pull/236

I had begun a refactor and forgot I needed these changes, so I incorporated them there.

PR #236, opened 2019-12-26
Split v1.py into smaller, more-specific route files

ryan-lane:
This change refactors v1.py route into smaller, more-specific route files, making it easier to navigate the code. It also reorders the imports to be properly ordered.

PR #237, opened 2019-12-26
Refactor diff view for services to match credentials diff

ryan-lane:
For consistency, this change refactors the controller and view for service diff to match how credentials is handled. This also adds a diff endpoint for services, with its own ACL.

Also, the diff views didn't show who edited the revisions, or when, so I've added modified_by and modified_date to both diffs.


ryan-lane:
This is based on https://github.com/lyft/confidant/pull/236, so the diff looks odd. Will rebase this, once 236 is merged.

PR #238, opened 2019-12-29
Use marshmallow for serializing responses

ryan-lane:
This change adds marshmallow support for serializing json responses in the API returns. The goal is to standardize responses, and to keep the ACL logic consistent in the output data.


confidant/schema/blind_credentials.py
```diff
@@ -0,0 +1,71 @@
...
+        ret = cls(
+            id=credential.id,
+            name=credential.name,
+            cipher_version=credential.cipher_version,
+            cipher_type=credential.cipher_type,
+            metadata=credential.metadata,
+            revision=credential.revision,
+            enabled=credential.enabled,
+            modified_date=credential.modified_date,
+            modified_by=credential.modified_by,
```
skiptomyliu: New to `attr`, but should we add a default for `data_key` in case the flag doesn't get passed in with `true`?  The others `credential_keys` and `credential_pairs` include defaults.

confidant/schema/blind_credentials.py
```diff
@@ -0,0 +1,71 @@
...
+from confidant.schema.auto_build_schema import AutobuildSchema
+
+
+@attr.s
+class BlindCredentialResponse(object):
+    id = attr.ib()
+    name = attr.ib()
+    cipher_version = attr.ib()
+    cipher_type = attr.ib()
+    data_key = attr.ib()
```
skiptomyliu: New to `attr`, but should we add a default for `data_key` in case the flag doesn't get passed in with `true`?  The others `credential_keys` and `credential_pairs` include defaults.

ryan-lane: Ah. Good catch. I just checked the model and data_key is a dict for blind_credentials, so yeah, this should have a dict default. Let me fix that.

skiptomyliu:
lgtm!

a lot cleaner with these changes

PR #245, opened 2020-01-07
Switch to a blueprint/create_app pattern

ryan-lane:
This change updates how the app is created for initialization, for easier testing. A change needed to switch to blueprints is to no longer use `app.config['<var>']`, but instead to use `settings.<var>`. This change also updates the encrypted settings and allows access directly to the encrypted settings from the settings variables.

### DIFF
--- a/confidant/app.py
+++ b/confidant/app.py
@@ -1,6 +1,7 @@
 import logging
 
 import boto3
+import guard
 from flask import Flask
 from flask_sslify import SSLify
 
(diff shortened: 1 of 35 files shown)


confidant/app.py
```diff
@@ -1,33 +1,62 @@
 import logging
 
 import boto3
+import guard
```
skiptomyliu: I think we need to update pip requirements for this

skiptomyliu: mmh nevermind

confidant/app.py
```diff
@@ -1,33 +1,62 @@
...
+def create_app():
+    static_folder = settings.STATIC_FOLDER
+
+    app = Flask(__name__, static_folder=static_folder)
+    app.config.from_object(settings)
+    app.config.update(settings.encrypted_settings.get_all_secrets())
+    app.debug = settings.DEBUG
+
+    if settings.SSLIFY:
+        sslify = SSLify(app, skips=['healthcheck'])  # noqa
```
skiptomyliu: does the var `sslify` need to be used elsewhere?  Looks like it's unreferenced at moment

ryan-lane: Ah, yeah, I could probably change this to just `SSLify(app, skips=['healthcheck'])` Let me update that.

PR #246, opened 2020-01-08
Switch from nose to pytest

ryan-lane:
This change switches the testting framework from nose to pytest, to make testing easier.


ryan-lane:
yep, but it pays off! :)

PR #248, opened 2020-01-10
Add permission hints to responses and update UI to use them

ryan-lane:
This change adds permissions hints for the UI, so that we can change the UI behavior based on the permissions available to the end-user.

### DIFF
--- a/confidant/routes/identity.py
+++ b/confidant/routes/identity.py
@@ -1,8 +1,11 @@
 from flask import blueprints, jsonify
 
 from confidant import authnz, settings
+from confidant.utils import misc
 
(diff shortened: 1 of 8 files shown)


confidant/routes/identity.py
```diff
@@ -32,6 +35,26 @@ def get_client_config():
     '''
     Get configuration to help clients bootstrap themselves.
     '''
+    permissions = {
```
skiptomyliu: ah nice, if I understand correctly, this should save us some RBAC API calls by storing these default permissions once the user logs in.

ryan-lane: @skiptomyliu yeah, these are the "global" permissions that apply to all views.

PR #249, opened 2020-01-10
Separate create service rbac check into its own call

skiptomyliu:
- Check that a user is able to create a service before calling the rbac check to determine if they're able to map credentials
- Small wording changes on mapping credential deny
- additional unit testing in credential manager

### DIFF
--- a/confidant/routes/services.py
+++ b/confidant/routes/services.py
@@ -313,6 +313,7 @@ def map_service_credentials(id):
         service.blind_credentials,
     )
     permissions = {
+        'create': True,
         'metadata': True,
         'get': True,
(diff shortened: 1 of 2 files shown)


confidant/routes/services.py
```diff
@@ -215,6 +215,18 @@ def map_service_credentials(id):
         revision = 1
         _service = None
 
+    if revision == 1 and not acl_module_check(
+          resource_type='service',
+          action='create',
+          resource_id=id,
+    ):
```
ryan-lane: Can you rebase the PR and update create in the permissions array for this?

PR #256, opened 2020-01-28
AWS ACM Private CA support

ryan-lane:
This change adds support for using ACM Private CA to provide an authenticated shim for issuing certificates for clients. Four endpoints are included, `GET /v1/certificates/<ca>/<cn>`, `POST /v1/certificates/<ca>`, `GET /v1/cas` and `GET /v1/cas/<ca>`. The GET certificate endpoint takes a certificate authority name (ca), and a common name argument (cn), generates a private key, CSR, and certificate for the given cn and returns a certificate, private_key and certificate_chain. The POST certificate endpoint takes a `<ca>` argument and a `csr` in the data body, and returns a certificate, and certificate chain issued for the CSR. Both endpoints accept a san argument (a URL param for GET, and a key/value in the POST body), which can be added as an extension. Both endpoints also accept a `validity` argument, which specify the number of days the certificate will be valid for. There's a setting to limit the max validity. The GET cas endpoint lists all configured CAs, and returns a certificate, certificate chain and a dictionary of tags for each CA. The GET ca endpoint returns certificate, certificate chain and a dictionary of tags for the specified ca.

This implementation support multiple CAs, defined in the `ACM_PRIVATE_CAS` setting, which is a comma delimited list of friendly names for the CAs. Settings for each CA are pulled from environment variables, postfixed with the friendly CA name, in all uppercase. e.g. `ACM_PRIVATE_CAS='testca'; ACM_PRIVATE_CA_ARN_TESTCA='arn:...'`

This implementation is just a pass-through, so each call will result in a certificate being generated, even if it's for the same CN with the same validity period. The idempotency token used in the ACM call is only effective for avoiding re-issuing certs within a short-period of time for the exact same CSR.

A simple cost-savings cache is included for the `GET /v1/certificates/<ca>/<cn>` endpoint, which will cache the csr/key/certificate in-memory, per-process. Locking is used between threads in the process to ensure that if a thread is issuing a certificate, other threads attempting to issue the same certificate will wait until the initial certificate issued, then will return the certificate from the cache.


confidant/services/certificatemanager.py
```diff
@@ -0,0 +1,199 @@
...
+            x509.NameAttribute(
+                NameOID.LOCALITY_NAME,
+                settings.LOCALITY_NAME
+            )
+        )
+    if settings.ACM_PRIVATE_CA_CSR_ORGANIZATION_NAME:
+        name_attributes.append(
+            x509.NameAttribute(
+                NameOID.ORGANIZATION_NAME,
+                settings.ORGANIZATION_NAME
```
skiptomyliu: looks like `ORGANIZATION_NAME` is missing from `settings`

skiptomyliu: err, possible just renamed to `ACM_PRIVATE_CA_CSR_ORGANIZATION_NAME`

confidant/services/certificatemanager.py
```diff
@@ -0,0 +1,199 @@
...
+            x509.NameAttribute(
+                NameOID.STATE_OR_PROVINCE_NAME,
+                settings.ACM_PRIVATE_CA_CSR_STATE_OR_PROVINCE_NAME
+            )
+        )
+    if settings.ACM_PRIVATE_CA_CSR_LOCALITY_NAME:
+        name_attributes.append(
+            x509.NameAttribute(
+                NameOID.LOCALITY_NAME,
+                settings.LOCALITY_NAME
```
skiptomyliu: `LOCALITY_NAME` -> `ACM_PRIVATE_CA_CSR_LOCALITY_NAME`

confidant/services/certificatemanager.py
```diff
@@ -0,0 +1,397 @@
...
+        From the provided csr object, return a list of the string values of the
+        subjust alternative name extension.
+        """
+        dns_names = []
+        try:
+            san = csr.extensions.get_extension_for_class(
+                x509.SubjectAlternativeName
+            )
+        except ExtensionNotFound:
+            san = []
```
skiptomyliu: nit:  I think better to set this to `None` instead of an empty list?    I'm interpreting the return to be a list, but looks like it's a single object where `san.value` is the list.

confidant/services/certificatemanager.py
```diff
@@ -0,0 +1,397 @@
...
+        if not item.response:
+            # Wait for response
+            while True:
+                # keep waiting while the lock is held.
+                if item.lock:
+                    logging.debug(
+                        'Sleeping in _get_cached_certificate_with_key'
+                        ' for {}'.format(cache_id)
+                    )
+                    time.sleep(.100)
```
skiptomyliu: ooc -- do you think it's worth adding a max loop iteration count as a fail safe to prevent an infinite loop if the cache fails to unlock?

confidant/services/certificatemanager.py
ryan-lane: Ah. Good catch. Will change.

confidant/services/certificatemanager.py
ryan-lane: gunicorn has a max timeout per-thread, but maybe it's a good idea to break before that, as I guess it's possible that not all wsgi workers do, and maybe someone's running this in a wsgi worker that isn't gunicorn.

confidant/authnz/rbac.py
```diff
@@ -1,27 +1,63 @@
...
+            # user.
+            for domain in domains:
+                match = cert_pattern.match(domain)
+                if not match:
+                    return False
+                service_name = match.group('service_name')
+                if not service_name:
+                    return False
+                if not authnz.user_is_service(service_name):
+                    return False
```
skiptomyliu: Do we need to add a check for `if service_name == authnzget_logged_in_user()`?  

Following the readme here:
https://github.com/lyft/confidant/pull/256/files#diff-d7018396c2fb69eeea73ddd1a3ce1e9cR590

confidant/services/certificatemanager.py
```diff
@@ -0,0 +1,449 @@
...
+
+    def issue_certificate_with_key(self, cn, validity, san=None):
+        """
+        Given the string common name, the validity length of the certificate (in
+        number of days), and a list of subject alternative names, return a dict
+        with the PEM encoded certificate, certificate chain, and private RSA
+        key.
+        """
+        with stats.timer('issue_certificate_with_key'):
+            cache_id = self.cache.get_cache_id(cn, validity, san)
```
skiptomyliu: Do we need to add a check that caching is enabled?  Basically checking: `ACM_PRIVATE_CA_CERTIFICATE_USE_CACHE_`

confidant/services/certificatemanager.py
```diff
@@ -0,0 +1,449 @@
...
+                except client.exceptions.RequestInProgressException:
+                    # Sleep for a maximum of 10 seconds
+                    if i >= 50:
+                        raise
+                    logging.debug(
+                        'Sleeping in get_certificate_from_arn for {}'.format(
+                            certificate_arn,
+                        )
+                    )
+                    time.sleep(.200)
```
skiptomyliu: perhaps dumb q -- with a maximum sleep of 10seconds, would it be possible to DOS confidant if theoretically ACM goes down or becomes degraded, and a bunch of services request certs where the processes then become tied up?

skiptomyliu:
Looks good, just a few minor comments.

confidant/authnz/rbac.py
ryan-lane: user_is_service does that: https://github.com/lyft/confidant/blob/master/confidant/authnz/__init__.py#L60-L74

confidant/services/certificatemanager.py
ryan-lane: Hm. I guess the setting isn't really correct right now as we're still caching the data, but not delivering it. I'm going to make a no-op cache class and set that if it's unused.

confidant/services/certificatemanager.py
ryan-lane: Hm. Yeah, that's definitely a possibility. It's a good concern. Maybe it's better to not sleep, but return a 429 here, which would cause the client to slow down and retry.

confidant/services/certificatemanager.py
skiptomyliu: no biggy, just wanted to call it out.  can always follow up on an additional PR.

ryan-lane: Fixed

confidant/services/certificatemanager.py
ryan-lane: Sending 429s with a Retry-After response header now

PR #261, opened 2020-02-12
Use ACL for both user and service checks

ryan-lane:
This change removes the special-cased code for controlling authz for services, and moves everything to using ACL checks. I've added a new default ACL function `default_acl`, which replicates the default behavior of service authz, while making it possible to fully apply ACL checks against services or users for all resources.

### DIFF
--- a/confidant/routes/services.py
+++ b/confidant/routes/services.py
@@ -72,41 +72,33 @@ def get_service(id):
         'update': False,
     }
     metadata_only = request.args.get('metadata_only', default=False, type=bool)
-    if authnz.user_is_user_type('service'):
-        if not authnz.user_is_service(id):
-            logging.warning('Authz failed for service {0}.'.format(id))
-            msg = 'Service is not authorized.'
-            return jsonify({'error': msg}), 401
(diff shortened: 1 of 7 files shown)


confidant/routes/services.py
```diff
@@ -72,41 +72,33 @@ def get_service(id):
...
-        permissions['metadata'] = acl_module_check(
-            resource_type='service',
-            action=action,
-            resource_id=id,
-        )
-        permissions['get'] = acl_module_check(
-            resource_type='service',
-            action=action,
-            resource_id=id,
+    logged_in_user = authnz.get_logged_in_user()
```
skiptomyliu: having trouble seeing it at moment, but if it's a service, will this return the service name?

skiptomyliu: nevermind, i think i see it, it will return the service/role name

https://github.com/lyft/confidant/blob/master/confidant/authnz/__init__.py#L191

ryan-lane: Yep! We set g.username whether it's user or service auth.

Issue #204, opened 2019-10-17
TypeError: a bytes-like object is required, not 'str'


jrosco:
Seeing the following error running Confidant version 5.2.0 using Docker image lyft/confidant:5.1.0

**Environment:**

Running build from Docker container
Running version 5.2.0
Using the following SAML Environment Vars set (these settings work in version 4)

USER_AUTH_MODULE=saml
SAML_SP_CERT_FILE=/srv/certificates/saml.sp.cert
SAML_SP_KEY_FILE=/srv/certificates/saml.sp.key
SAML_SECURITY_SLO_RESP_SIGNED=false
SAML_SECURITY_ASSERTIONS_SIGNED=true
SAML_IDP_CERT_FILE=/srv/certificates/saml.idp.cert
SAML_SECURITY_MESSAGES_SIGNED=false

Seeing a type error `TypeError: a bytes-like object is required, not 'str' Error `
(see below log)
```
ERROR:confidant.app:Exception on /v1/login [GET]
Traceback (most recent call last):
  File "/venv/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/venv/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/venv/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/venv/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/venv/lib/python3.6/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/venv/lib/python3.6/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/srv/confidant/confidant/routes/v1.py", line 38, in login
    return authnz.log_in()
  File "/srv/confidant/confidant/authnz/__init__.py", line 139, in log_in
    return user_mod.log_in()
  File "/srv/confidant/confidant/authnz/userauth.py", line 558, in log_in
    return flask.redirect(self.login_redirect_url(return_to=self_page))
  File "/srv/confidant/confidant/authnz/userauth.py", line 816, in login_redirect_url
    auth = self._saml_auth()
  File "/srv/confidant/confidant/authnz/userauth.py", line 723, in _saml_auth
[...]

PR #268, opened 2020-02-25
Decode private key when loaded from file as base64

ryan-lane:
Python3 fix for SAML auth when using a client key.


ryan-lane:
Fixes #204

PR #269, opened 2020-03-13
Api docs

ryan-lane:


ryan-lane:
@skiptomyliu pushed this in to save my place. going to have a bunch more coming in this PR :D

PR #270, opened 2020-03-14
Add credential rotation support

irhkang:


confidant/models/credential.py
```diff
@@ -142,6 +160,40 @@ def _get_decrypted_credential_pairs(self):
         _credential_pairs = json.loads(_credential_pairs)
         return _credential_pairs
 
+    @property
+    def next_rotation_date(self):
+        """
+        Return when a credential needs to be rotated for security purposes.
+        """
+        if not self.requires_rotation or self.rotation_frequency is None:
```
irhkang: Short circuiting on `self.rotation_frequency is None` is risky in that if the env dict isn't present, we might not rotate.  I think it's worth doing this to be generic but I would expect other callers to ensure that the env var is present

confidant/models/credential.py
```diff
@@ -142,6 +160,40 @@ def _get_decrypted_credential_pairs(self):
...
+            return None
+
+        # If a credential has never been rotated or been decrypted,
+        # immediately rotate
+        if self.last_rotation_date is None:
+            return datetime.utcnow()
+
+        if self.last_decrypted_date and \
```
ryan-lane: Use parens, rather than backslashes for multi-line

confidant/models/credential.py
```diff
@@ -52,6 +52,11 @@ class Meta:
     modified_by = UnicodeAttribute()
     documentation = UnicodeAttribute(null=True)
 
+    # Classification info (eg: FINANCIALLY_SENSITIVE)
+    category = UnicodeAttribute(null=True)
```
ryan-lane: Can we do:

```
tags = ListAttribute(default=list)
```

confidant/models/credential.py
```diff
@@ -142,6 +160,40 @@ def _get_decrypted_credential_pairs(self):
...
+
+        return self.last_rotation_date + timedelta(days=self.rotation_frequency)
+
+    @property
+    def rotation_frequency(self):
+        """
+        Return how frequently a credential needs to be rotated
+        Return None if category is not in CREDENTIAL_ROTATION_DAYS config
+        """
+        return settings.CREDENTIAL_ROTATION_DAYS.get(self.category)
```
ryan-lane: For this, let's iterate over the list and take the lowest value:

```
days = settings.MAXIMUM_ROTATION_DAYS
for tag in self.tags:
    tag_days = settings.CREDENTIAL_ROTATION_DAYS.get(tag)
    if tag_days < days:
        days = tag_days
return days
```

confidant/models/credential.py
```diff
@@ -142,6 +160,40 @@ def _get_decrypted_credential_pairs(self):
...
+
+    @property
+    def requires_rotation(self):
+        """
+        Some credentials need to be periodically rotated
+        """
+        return self.category == settings.FINANCIALLY_SENSITIVE
```
ryan-lane: For this, you don't want to go off a specific tag, you want to go off `self.next_rotation_date`

confidant/settings.py
```diff
@@ -576,6 +576,10 @@ def str_env(var_name, default=''):
     ACM_PRIVATE_CA_SETTINGS[ca] = ca_settings
 
 
+FINANCIALLY_SENSITIVE = 'FINANCIALLY_SENSITIVE'
```
ryan-lane: Don't think we want a hardcoded setting like this

confidant/models/credential.py
skiptomyliu: Oh wait, I thought we were going to add tags in settings.py or as env vars?

confidant/models/credential.py
```diff
@@ -93,6 +101,17 @@ def diff(self, other_cred):
                 'added': new.documentation,
                 'removed': old.documentation
             }
+        if set(old.tags) != set(new.tags):
+            diff['tags'] = {
+                'added': new.tags,
+                'removed': old.tags
+            }
+        if old.last_rotation_date != new.last_rotation_date:
```
irhkang: note: deliberately not putting last_decrypted_date in the diff

confidant/models/credential.py
```diff
@@ -142,6 +161,41 @@ def _get_decrypted_credential_pairs(self):
...
+                days = rotation_days
+        return self.last_rotation_date + timedelta(days=days)
+
+    @property
+    def exempt_from_rotation(self):
+        """
+        Credentials with certain tags (say financial credential)
+        must be rotated.  Credentials without these tags can be exempt.
+        """
+        return len(set(self.tags) & set(settings.TAGS_REQUIRING_ROTATION)) == 0
```
irhkang: I much prefer this to a ROTATION_EXEMPT tag as people won't need to migrate all records - they can just tag the few they care about

confidant/models/credential.py
```diff
@@ -93,6 +101,17 @@ def diff(self, other_cred):
                 'added': new.documentation,
                 'removed': old.documentation
             }
+        if set(old.tags) != set(new.tags):
+            diff['tags'] = {
+                'added': new.tags,
+                'removed': old.tags
+            }
```
ryan-lane: For this you need to do a list diff for added/removed. added should only contain fields that were actually added, and removed should only contain fields actually removed. We only display the diff for the field if things changed.

irhkang: Yea good point

confidant/models/credential.py
```diff
@@ -142,6 +161,41 @@ def _get_decrypted_credential_pairs(self):
...
+            return datetime.utcnow()
+
+        days = settings.MAXIMUM_ROTATION_DAYS
+        for tag in self.tags:
+            rotation_days = settings.ROTATION_DAYS_CONFIG.get(tag)
+            if rotation_days is None:
+                continue
+            if days is None or rotation_days < days:
+                days = rotation_days
+        return self.last_rotation_date + timedelta(days=days)
```
ryan-lane: days can be None here still. I'd imagine we want the behavior for this to not flag a credential for rotation if settings.MAXIMUM_ROTATION_DAYS isn't set, or if no tags are applied?

irhkang: Yea I'm aware of this - I wanted to specify this in the readme.  I could make a check for that if you feel strongly

```
if self.exempt_from_rotation or settings.MAXIMUM_ROTATION_DAYS is None:
    return None
```

ryan-lane: Code should always be defensive. If the default for `settings.MAXIMUM_ROTATION_DAYS` is None, this is going to throw an exception:

```
>>> from datetime import timedelta
>>> timedelta(days=None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported type for timedelta days component: NoneType
```

ryan-lane: Oh, it's using int_env, so I guess the default is 0. This would mean all credentials would always show as needing rotation, but I guess if people aren't using the feature, it's not an issue.

irhkang: Yea it shouldn't cause problems if people aren't using the feature and the README will say very explicitly that MAXIMUM_ROTATION_DAYS must be set

tests/unit/confidant/models/credential_test.py
```diff
@@ -43,6 +43,28 @@ def test_not_equals(mocker):
     assert cred1.equals(cred2) is False
 
 
+def test_not_equals_different_tags(mocker):
```
skiptomyliu: should we add a test to check for tags equal?

irhkang: Sure - I modified an existing test

PR #271, opened 2020-03-14
Add maintenance script and pynamodb model for archiving and restoring credentials

ryan-lane:
This change adds the ability to archive disabled credentials that haven't been modified in a specified set of days into a separate dynamodb archive table.

Note that this does not remove references to archived credentials in old service mappings. `Credential.batch_get` will silently filter out any invalid or missing IDs, and every call that fetches credentials from the service record uses `batch_get`. On service update, or when reverting to a revision, we now filter the list of IDs prior to saving. We filter these after the RBAC check, though, so it's necessary for RBAC modules to handle NotFound errors, if they're doing `Credential.get` from the list of IDs, rather than using `Credential.batch_get`.


confidant/models/credential.py
```diff
@@ -145,3 +148,30 @@ def _get_decrypted_credential_pairs(self):
     @property
     def decrypted_credential_pairs(self):
         return(self._get_decrypted_credential_pairs())
+
+
+class CredentialArchive(CredentialBase):
```
irhkang: You might want an archive_at datetime field for auditing purposes

confidant/routes/services.py
```diff
@@ -285,12 +285,19 @@ def map_service_credentials(id):
         except keymanager.ServiceCreateGrantError:
             msg = 'Failed to add grants for {0}.'.format(id)
             logging.error(msg)
+    credentials = credentialmanager.get_credentials(data.get('credentials'))
```
irhkang: this probably won't happen but does any service have >100 credentials?  The docs say that BatchGet returns at most 100 items

irhkang: also a nit - might be worth making this into a function or adding a comment.  Otherwise it's not obvious what's going on behind the scenes

```
def filter_archived_credentials(credentials):
    filtered_credentials = credentialmanager.get_credentials(data.get('credentials')). # BatchGetItem will automatically remove credentials that don't exist
    return [f.id for f in filtered_credentials]
```

confidant/scripts/archive.py
```diff
@@ -0,0 +1,93 @@
...
+        Option('--force', action='store_true', dest='force', default=False),
+    ]
+
+    def run(self, days, force):
+        if not settings.DYNAMODB_TABLE_ARCHIVE:
+            logger.error('DYNAMODB_TABLE_ARCHIVE is not configured, exiting.')
+            return 1
+        for credential in Credential.data_type_date_index.query('credential'):
+            tz = credential.modified_date.tzinfo
+            now = datetime.now(tz)
```
irhkang: why not use `utcnow()`? modified date is a UTC timestamp.  That saves you from all the timezone conversions

confidant/scripts/archive.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class ArchiveCredentials(Command):
+
+    option_list = [
+        Option('--days', dest='days', required=True, type=int),
+        Option('--force', action='store_true', dest='force', default=False),
+    ]
+
+    def run(self, days, force):
```
irhkang: you might want to consider adding a sleep or something to make sure you don't blow past your WCU

skiptomyliu:
> so it's necessary for RBAC modules to handle NotFound errors, if they're doing Credential.get from the list of IDs, rather than using `Credential.batch_get`

Making sure I understand - the UI will show "[cred] (archived)" -- this means the credential will still be clickable, in this case we want RBAC to check if a credential is disabled and return denied, that correct?

ryan-lane:
It will not be clickable, because it'll be archived and no longer in the
table. In the RBAC plugin, we need to catch DoesNotExist records and ignore
them.

confidant/routes/services.py
ryan-lane: True. It's unlikely any have >100, but we should probably do a check to see if there's unprocessed records, and if so, fetch those as well. Not sure if pynamo handles that for you or not. In either case, it's out of scope for this PR, since it's a general problem in the software.

ryan-lane: We need to use the credentials again later in the function (for the return). I just moved the fetching of them up so that I could also get a list of filtered IDs (that's documented).

ryan-lane: Docs say that pynamo handles batch limits invisibly: https://pynamodb.readthedocs.io/en/latest/batch.html

confidant/scripts/archive.py
ryan-lane: This is a good point. I think I'm going to make a function for deleting and saving that will catch exceptions for provisioned write and will backoff and retry.

confidant/routes/services.py
irhkang: I saw that doc but the code doesn't seem like it uses UnprocessedKeys https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/connection/base.py#L1112.

If we don't have any service having >100 keys and the chances of one having >100 in the future is small , it might be worth just adding a stat and alarming on the stat.  You can implement pagination when needed.

confidant/scripts/archive.py
ryan-lane: Tried this: ``can't subtract offset-naive and offset-aware datetimes``

irhkang: Yikes it looks like utcnow doesn't actually populate the tzinfo field (https://stackoverflow.com/questions/2331592/why-does-datetime-datetime-utcnow-not-contain-timezone-information).  Ignore the comment - apologies for the noise

confidant/routes/services.py
ryan-lane: Hm. That's good to know. I'm not going to work that into this PR, but will address this in a followup, since we need to handle this case everywhere, not just the script. I'll probably add a wrapper function for batch gets that handles this, that we call for for all batch gets.

confidant/models/credential.py
ryan-lane: Good point. Let me add that.

confidant/routes/services.py
ryan-lane: It does; it's in the model class: https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/models.py#L321-L332

confidant/scripts/archive.py
```diff
@@ -0,0 +1,169 @@
...
+                continue
+            saves = []
+            deletes = []
+            # save the current record.
+            archive_credential = CredentialArchive.from_credential(
+                credential,
+            )
+            saves.append(archive_credential)
+            # fetch and save every revision
+            _range = range(1, credential.revision + 1)
```
irhkang: thoughts on putting the next few 5 lines into a method in credentialmanager? Because I was going to do that myself for https://github.com/lyft/confidant/pull/272

confidant/scripts/archive.py
```diff
@@ -0,0 +1,169 @@
...
+                archive_revision = CredentialArchive.from_credential(
+                    revision,
+                )
+                saves.append(archive_revision)
+                deletes.append(revision)
+            deletes.append(credential)
+            try:
+                self.save(saves, force=force)
+            except Exception:
+                logger.exception(
```
irhkang: If we're productionizing this we should add a stat here to alarm on

confidant/scripts/archive.py
```diff
@@ -0,0 +1,169 @@
...
+                        'Skipping enabled credential {}'.format(credential.id)
+                    )
+                    continue
+                credentials.append(credential)
+        else:
+            for credential in Credential.data_type_date_index.query(
+                'credential'
+            ):
+                tz = credential.modified_date.tzinfo
+                now = datetime.now(tz)
```
irhkang: Not necessary but a slight optimization would be to move these 2 lines outside the loop since tzinfo is always UTC time (thanks to pynamo)

confidant/scripts/restore.py
```diff
@@ -0,0 +1,134 @@
...
+        if ids and _all:
+            logger.error('--ids and --all arguments are mutually exclusive')
+            return 1
+        if not ids and not _all:
+            logger.error('Either --ids or --all argument must be provided')
+            return 1
+        if ids:
+            # filter strips an empty string
+            _ids = [_id.strip() for _id in list(filter(None, ids.split(',')))]
+            if not _ids:
```
irhkang: nit: check for len > 100  for pagination issues?

irhkang:
Also you could add the salt configs here. Not necessary though - we could do that in a follow up PR

confidant/models/credential.py
```diff
@@ -29,19 +29,19 @@ class Meta:
...
-        if settings.DYNAMODB_URL:
-            host = settings.DYNAMODB_URL
-        region = settings.AWS_DEFAULT_REGION
-        connection_cls = DDBConnection
-        session_cls = DDBSession
+        projection = AllProjection()
+        read_capacity_units = 10
+        write_capacity_units = 10
+    data_type = UnicodeAttribute(hash_key=True)
+    modified_date = UTCDateTimeAttribute(range_key=True)
```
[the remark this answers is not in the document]
ryan-lane: Keeps things simpler to use the same index in both places, as any orchestration is copy/paste. We fetch every record and aren't sorting by the time anyway :)

confidant/scripts/archive.py
ryan-lane: Ah. Was assuming this would be run by hand, but I guess I didn't consider that we could possible run this in a cron... I'll add a stat.

confidant/scripts/restore.py
ryan-lane: No pagination issues. batch_get handles it. I linked to the implementation in another comment somewhere.

confidant/scripts/archive.py
ryan-lane: Yep. Can do.

PR #272, opened 2020-03-17
Update last_rotation_date field when a credential is changed

irhkang:
Note: branching off of https://github.com/lyft/confidant/pull/270 as that has the field needed

There are 3 cases here

1. Creating a credential: always set last_rotation_date to the current time
2. Updating a credential: if the credential key has changed, update the last_rotation_date to the current time. Otherwise leave it alone
3. Reverting a credential: use the last_rotation_date of the revert credential.  We have logic elsewhere to rotate the credential even if the revert doesn't have this field populated

We deliberately do not send the last_rotation_date back to the client


confidant/routes/credentials.py
```diff
@@ -370,7 +374,9 @@ def update_credential(id):
             }
             return jsonify(ret), 400
         update['credential_pairs'] = json.dumps(credential_pairs)
+        update['last_rotation_date'] = datetime.utcnow()
```
irhkang: There is a key assumption I'm making here that I would like someone to validate  - I'm assuming that if `credential_pairs` is in the payload (`data`), then it has changed.

However, I have looked at the code and I see that it doesn't explicitly check if the key has changed compared to it's previous revisions  - it just checks for conflicts with other keys.  If we think this is a risk, I can fetch all the revisions but I thought that might be overkill..

ryan-lane: If credential_pairs is in the payload data, it's changed, yeah. That's the only time we should change the rotation time.

ryan-lane: Though... the client sends in a full object, so I think it's really necessary to compare the values, and only adjust the rotation time if they've changed.

confidant/routes/credentials.py
```diff
@@ -514,6 +522,7 @@ def revert_credential_to_revision(id, to_revision):
             cipher_version=revert_credential.cipher_version,
             modified_by=authnz.get_logged_in_user(),
             documentation=revert_credential.documentation,
+            last_rotation_date=revert_credential.last_rotation_date,
```
ryan-lane: For revert we also need to check whether or not the credentials differ. If someone reverts a credential, but it only changes the name or tags, for instance, the rotation time hasn't changed.

irhkang: But the rotation time of the credential should have already reflected that.  For instance:

```
version 1: (credential_pair: "asdf", last_rotation_date: T0, name: "the OG key")

# credential_pair changed so update last_rotation_date
version 2: (credential_pair: "bbbb", last_rotation_date: T1, name: "new key). 

# credential_pair is the same so don't update last_rotation_date
version 3: (credential_pair: "bbbb", last_rotation_date: T1, name: "still the new key")
```

Say version 3 is the current version and we want to revert back to version 2, `last_rotation_date: T1` is still accurate and *it's accurate because of our checks in the update credential endpoint*.  

The update endpoint essentially creates an invariant where `last_rotation_date` reflects the time the key was actually changed, not when the object was updated.  This endpoint just takes advantage of that invariant.

confidant/routes/credentials.py
irhkang: >Though... the client sends in a full object, so I think it's really necessary to compare the values, and only adjust the rotation time if they've changed.

What does this mean? It's not referring to  checking that `credential_pairs` is in the payload right?

irhkang:
Also you could add the salt configs here.  Not necessary though - we could do that in a follow up PR

confidant/routes/credentials.py
ryan-lane: Ah. Good point.

confidant/routes/credentials.py
ryan-lane: The UI modifies the current object in place, then sends the entire thing in to the update endpoint. That said, it doesn't matter the behavior anyway, because you want to make sure the credentials have actually changed, because they could just send the same value in to fool your rotation detection.

confidant/routes/credentials.py
```diff
@@ -804,19 +819,17 @@ def update_credential(id):
                 'conflicts': conflicts
             }
             return jsonify(ret), 400
+
+        # If the credential pair passed in the update is different from the
+        # decrypted credential pair of the most recent revision, assume that
+        # this is a new credential pair and update last_rotation_date
+        if credential_pairs != _cred.decrypted_credential_pairs:
```
irhkang: It is possible to enter this condition if you just change the key name while leaving the token the same - we deliberately ignore this case

confidant/routes/credentials.py
```diff
@@ -765,22 +768,34 @@ def update_credential(id):
...
-        _cred.revision
-    )
-    update['name'] = data.get('name', _cred.name)
+    if not isinstance(data.get('metadata', {}), dict):
+        return jsonify({'error': 'metadata must be a dict'}), 400
+
+    update = {
+        'name': data.get('name', _cred.name),
+        'last_rotation_date': _cred.last_rotation_date,
+        'credential_pairs': _cred.credential_pairs,
```
irhkang: changing to use the encrypted string rather than the decryped dict as Ciphermanager.encrypt expects a string

ryan-lane: That's going to be an encrypted string. You're going to double encrypt here.

ryan-lane: you want: `json.dumps(_cred.decrypted_credential_pairs)`

ryan-lane: alternatively, you can keep the encrypted value here, and move the part that does encryption into the if section and only apply it to what's in the data field

irhkang: Ah you're right

PR #273, opened 2020-03-19
Use module local logger consistently across codebase

ryan-lane:


irhkang:
Perhaps this is too much of a tangent but this PR is a decent place to start using python 3 format strings. 

Instead of `'{0} get credential {1}'.format(authnz.get_logged_in_user(), id)`, we could do `f'{authnz.get_logged_in_user()} get credential {id}'`

ryan-lane:
I'll be making a bunch of cleanup PRs in the next week or two, so don't want to lump too much together in a single PR if possible :)

PR #274, opened 2020-03-19
Add metadata_only flag to GET /v1/credentials/<ID>

irhkang:
This PR does 2 things.

1. Add a metadata_only flag to GET /v1/credentials/<ID>.  This provides more granular controls for fetching a credential.  One benefit this has is that it's very easy to tell when a credential_pair is sent in the response (when metadata_only=False). This gives us an easier way to see when a credential was last decrypted

2. Update the credential's last decrypted time if the credential_pairs was sent in the response. **NOTE: this is opt in - there is a env variable that must be set for the save to occur**. We make this opt in to not disturb existing users


confidant/routes/credentials.py
```diff
@@ -187,16 +194,22 @@ def get_credential(id):
...
-    if acl_module_check(resource_type='credential',
-                        action='get',
-                        resource_id=id):
+    if not metadata_only and acl_module_check(resource_type='credential',
+                                              action='get',
+                                              resource_id=id):
         permissions['get'] = True
         include_credential_pairs = True
+
+        if settings.ENABLE_SAVE_LAST_DECRYPTION_TIME:
```
irhkang: Make this opt in - the setting defaults to false

confidant/models/credential.py
```diff
@@ -184,11 +184,11 @@ def next_rotation_date(self):
         # If a credential has never been rotated or been decrypted,
         # immediately rotate
         if self.last_rotation_date is None:
-            return datetime.utcnow()
+            return datetime.now()
```
irhkang: now() has the timezone so we don't do subtraction between tz aware and tz unaware datetimes

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+## 6.4.0
+
+* This release adds support for 2 small features:
+
+  * Add a `metadata_only` param to `GET /v1/credentials/<ID>`. For instance, if the request is
+    `GET /v1/credentials/123?metadata_only=true`, the response will not contain the credential_pairs.
+    `metadata_only` defaults to `false` so that it is backwards compatible. The purpose of this
+    is to give users finer controls when deciding whether to send back `credential_pairs`.
+  * Automatically update the `last_decrypted_date` on a credential when the `credential_pairs` are
+    sent back to the client. Sending a credential_pair to the client implies that a credential has been
```
skiptomyliu: nit minor formatting: "credential pair" or `credential_pair`

PR #275, opened 2020-03-19
Save archived credential when updating last_decrypted_date

irhkang:
This fixes 2 things

- `metadata_only` should be parsed with `flask_restful.inputs.boolean`, not `bool` as `bool()` evaluates everything to True.  The test for this has changed and now can catch this.

- When we update the last_decrypted_date of a credential, we also update that field on the archive credential model.  

We also limit IDs for `GET /v1/credentials/<ID>` to be of data type `credential` (as opposed to `credential` and `archive-credential`)


confidant/routes/credentials.py
```diff
@@ -160,7 +161,7 @@ def get_credential(id):
     metadata_only = request.args.get(
         'metadata_only',
         default=False,
-        type=bool,
+        type=inputs.boolean,
```
irhkang: See https://github.com/flask-restful/flask-restful/issues/397.  

Oddly, I don't remember running into this issue when using `reqparse` so I assume it's just because I'm using `request.args.get`.  Either way, the bug is definitely fixed now

confidant/routes/credentials.py
```diff
@@ -6,6 +6,7 @@
 
 from datetime import datetime
 from flask import blueprints, jsonify, request
+from flask_restful import inputs
```
ryan-lane: Please don't add flask_restful, it's deprecated, and we're using blueprints for everything.

confidant/routes/credentials.py
ryan-lane: I don't see how this is an issue. We're not seeing this for the get_service endpoint.

confidant/routes/credentials.py
```diff
@@ -201,8 +202,23 @@ def get_credential(id):
         include_credential_pairs = True
 
         if settings.ENABLE_SAVE_LAST_DECRYPTION_TIME:
-            credential.last_decrypted_date = datetime.now()
+            now = datetime.now()
+            credential.last_decrypted_date = now
             credential.save()
+            # Also try to save the archived credential if ID
+            # corresponds to a 'credential'
+            if credential.data_type == 'credential':
```
ryan-lane: Let's actually just limit this endpoint to credential.data_type == 'credential' up in line 184

irhkang: that causes potential backwards incompatibilities with existing users. 

It's not a huge deal in this case though

confidant/routes/credentials.py
irhkang: Spoke offline - moving to a helper function

confidant/routes/credentials.py
```diff
@@ -201,8 +196,21 @@ def get_credential(id):
...
         if settings.ENABLE_SAVE_LAST_DECRYPTION_TIME:
-            credential.last_decrypted_date = datetime.now()
+            now = datetime.now()
+            credential.last_decrypted_date = now
             credential.save()
+            # Also try to save the archived credential to stay consistent
+            try:
+                archived_credential = Credential.get(
+                    '{}-{}'.format(id, credential.revision)
+                )
```
ryan-lane: Put both the saves in here, and add exception handling for the save failing in the list of exceptions.

confidant/routes/credentials.py
```diff
@@ -201,8 +196,21 @@ def get_credential(id):
...
+            except DoesNotExist:
+                logger.warning('Archived credential {}-{} not found'.format(
+                        id, credential.revision)
+                )
+                return jsonify({}), 404
```
ryan-lane: I don't think we should return 404 here. I think we should just continue.

confidant/routes/credentials.py
```diff
@@ -201,8 +196,21 @@ def get_credential(id):
...
             credential.save()
+            # Also try to save the archived credential to stay consistent
+            try:
+                archived_credential = Credential.get(
+                    '{}-{}'.format(id, credential.revision)
+                )
+            except DoesNotExist:
+                logger.warning('Archived credential {}-{} not found'.format(
+                        id, credential.revision)
+                )
```
ryan-lane: I'd make this an error.

PR #276, opened 2020-03-20
Update docker-compose to make a full environment, with kms and auth

ryan-lane:
This change updates the docker-compose setup, and improves the local developer experience. The new docker-compose setup includes a local kms server, and a local saml idp. It's has a full confidant configuration that should relatively well replicate a somewhat production looking environment.


confidant/services/keymanager.py
```diff
@@ -10,35 +10,47 @@
...
+
+def _get_auth_kms_client():
+    return confidant.clients.get_boto_client(
+        'kms',
+        config={'name': 'keymanager', 'config': _get_boto_config()},
+        endpoint_url=settings.KMS_URL,
+    )
+
+
+def _get_at_rest_kms_client():
```
skiptomyliu: This looks to duplicate `_get_auth_kms_client()`, do we need need an additional variable in `settings.py`, where one is for auth and the other is for the at rest client?

ryan-lane: Oh, interestingly enough, back in [this change](https://github.com/lyft/confidant/commit/d7ccb71d4b1e7d02f8c5cb9e691bf7807ecb1b10#diff-929092871d7658f82a1785ab913e6e47R16-R28) I meant to split these into two clients, but didn't give them separate names, so they actually reference the same client. Let me update this to split them into two clients (so they have their own connection pools).

PR #279, opened 2020-03-27
Only load credential pairs in UI if user asks to unmask or edits

ryan-lane:
This change improves the auditing and last-viewed behavior of confidant by only fetching decrypted credential pairs when necessary for the UI. If the end-user loads a credential, it loads `metadata_only=True` by default, and if they unmask the credentials, or edit the credential, it'll fetch the credential with `metadata_only=False`, fetching the credential pairs, and replacing them in-place in the view. The view will keep them loaded while in the same credential view, and will not re-fetch the credential for unmask/edit, as long as the user doesn't leave and re-visit the page.

### DIFF
--- a/confidant/models/credential.py
+++ b/confidant/models/credential.py
@@ -53,7 +53,7 @@ class CredentialBase(Model):
     modified_by = UnicodeAttribute()
     documentation = UnicodeAttribute(null=True)
     # Classification info (eg: FINANCIALLY_SENSITIVE)
-    tags = ListAttribute(default=list)
+    tags = ListAttribute(default=list, null=True)
     last_decrypted_date = UTCDateTimeAttribute(null=True)
     last_rotation_date = UTCDateTimeAttribute(null=True)
 
--- a/confidant/models/credential.py
+++ b/confidant/models/credential.py
@@ -53,7 +53,7 @@ class CredentialBase(Model):
     modified_by = UnicodeAttribute()
     documentation = UnicodeAttribute(null=True)
     # Classification info (eg: FINANCIALLY_SENSITIVE)
-    tags = ListAttribute(default=list, null=True)
+    tags = ListAttribute(default=list)
     last_decrypted_date = UTCDateTimeAttribute(null=True)
     last_rotation_date = UTCDateTimeAttribute(null=True)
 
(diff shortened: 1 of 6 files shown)


confidant/models/credential.py
```diff
@@ -53,7 +53,7 @@ class CredentialBase(Model):
     modified_by = UnicodeAttribute()
     documentation = UnicodeAttribute(null=True)
     # Classification info (eg: FINANCIALLY_SENSITIVE)
-    tags = ListAttribute(default=list)
+    tags = ListAttribute(default=list, null=True)
```
ryan-lane: This change is necessary because otherwise it's impossible to create new credentials, as the UI doesn't pass this value in yet. In general, though, it's probably the behavior we want.

irhkang: You don't need this as default will take care of it - whether you're re-saving a credential or creating a new one, it'll create an empty list as the default value.  Setting it to null creates additional problems as you would need to change https://github.com/lyft/confidant/blob/master/confidant/models/credential.py#L194 to handle the `None` case

I also verified on a onebox copying a test fixture (the only change I made was changing the type of `data_key` to be binary):

```
>>> from confidant.models.credential import Credential
>>> a = Credential(
...     id='1234',
...     revision=1,
...     data_type='credential',
...     enabled=True,
...     name='Test credential',
...     credential_pairs='akjlaklkaj==',
...     data_key=b'slkjlksfjklsdjf==',
...     cipher_version=2,
...     metadata={},
...     modified_date=datetime.now(),
...     modified_by='test@example.com',
...     documentation='',
... )
>>> a.save()
{'ConsumedCapacity': {'CapacityUnits': 1, 'TableName': 'confidant-development-iad'}}
>>> b = Credential.get('1234')
>>> b.tags
[]
```

skiptomyliu: I think it's because of https://github.com/lyft/confidant/blob/master/confidant/routes/credentials.py#L600-L614

pynamodb is restricting passing in `None`

ryan-lane: From what I remember, null=true allows pynamo to save nulls (like empty lists, which to dynamo is a null). The default ensures that if you load it, and it's null, it'll be a list.

I added this because I was getting stacktraces when trying to create credentials in dev.

ryan-lane: Verified that the model works correctly with a None saved tags:

```
>>> from confidant.models.credential import Credential
>>> c = Credential.get('92bd1bfe74aa48929facd76fe0d87df7')
>>> c.next_rotation_date
datetime.datetime(2020, 4, 6, 7, 49, 51, 703224, tzinfo=datetime.timezone.utc)
```

ryan-lane: Updated to not pass nulls in, and removed null=true from the model

PR #281, opened 2020-03-28
Python 3 bytes vs string fixes for bootstrap script

ryan-lane:
Fixes for bytes vs strings issue in bootstrap script. This change also makes some fixes to how the KMS client is created and passed around, for proper testing in the docker-compose environment.

### DIFF
--- a/confidant/encrypted_settings.py
+++ b/confidant/encrypted_settings.py
@@ -4,6 +4,8 @@ import logging
 import json
 
 from cryptography.fernet import Fernet
+
+import confidant.clients
 from confidant.lib import cryptolib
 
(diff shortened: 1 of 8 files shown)


ryan-lane:
@tristans-stripe I saw your earlier PR, and investigated this. I think this should address your issue!

tristans-stripe:
Hey @ryan-lane - thanks for this quick fix! Apologies for the malformed PR earlier. Looks like this PR should fix the issue we were seeing.

Thanks again.

ryan-lane:
Great! No worries, thanks for bringing it to my attention :)

PR #284, opened 2020-04-01
Fix links between documents in the docs

ryan-lane:
When we switched from middleman to sphinx, we forgot to fix some inter-doc links. This change updates the links to point to the rendered html files as relative links.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@
   the CA in AWS.
 
   The implementation supports multiple CAs. For configuration information, see [the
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -68,7 +68,7 @@
   the CA in AWS.
 
   The implementation supports multiple CAs. For configuration information, see [the
-  certificate authority settings section in the docs](https://lyft.github.io/confidant/configuration.html#certificate-authority-settings).
+  certificate authority settings section in the docs](configuration.html#certificate-authority-settings).
```
irhkang, line 71: Err am I missing something? These links don't work for me - even when I use master https://github.com/lyft/confidant/blob/master/CHANGELOG.md

PR #286, opened 2020-04-02
Add tags to the edit/new credential UI

ryan-lane:
This change adds tags to the edit and new credential UI. The UI currently only allows selection of server defined tags, and specifically only ones defined by the credential rotation settings. Future iterations should allow more server defined tags, or allow user input that isn't from a server defined list.

### DIFF
--- a/confidant/public/modules/resources/controllers/CredentialDetailsCtrl.js
+++ b/confidant/public/modules/resources/controllers/CredentialDetailsCtrl.js
@@ -257,7 +257,12 @@
                         $scope.credential.mungedTags.splice(i, 1);
                         continue;
                     }
+                    // strip duplicates
                     if (tagItem.id in _credential.tags) {
+                        continue;
+                    }
+                    // strip empty tag selection
+                    if (tagItem.id === '') {
                         continue;
                     }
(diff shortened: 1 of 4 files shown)


confidant/public/modules/resources/controllers/CredentialDetailsCtrl.js
```diff
@@ -228,6 +251,22 @@
...
+                    }
+                    // strip duplicates
+                    if (tagItem.id in _credential.tags) {
+                        continue;
+                    }
+                    // strip empty tag selection
+                    if (tagItem.id === '') {
+                        continue;
+                    }
+                    _credential.tags.push(tagItem.id);
```
skiptomyliu, line 268: Should we use a set?  I think we're able to add multiple duplicate tags if we wanted to.

ryan-lane: I'm stripping them in line 261

skiptomyliu: ah thx, missed that

PR #287, opened 2020-04-03
Show next rotation time in credential UI

ryan-lane:
Add info in the credential UI for the next required rotation date for a credential.

### DIFF
--- a/confidant/public/modules/resources/views/credential-details.html
+++ b/confidant/public/modules/resources/views/credential-details.html
@@ -53,10 +53,10 @@
     </div>
     <div class="form-group" ng-show="credential.permissions.get || globalPermissions.credentials.create">
       <label for="credentialPairInputs">Credential Pairs
-        <span class="glyphicon glyphicon-lock"></span>
+        <span class="glyphicon glyphicon-lock" data-toggle="tooltip" title="This attribute is encrypted at-rest"></span>
         <span ng-show="!editableForm.$visible">
-          <a ng-click="toggleCredentialMask()" ng-show="!showCredentialPairs && !editableForm.$visible"><span class="glyphicon glyphicon-eye-open"></span></a>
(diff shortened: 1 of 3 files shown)


ryan-lane:
Screen Shot 2020-04-02 at 8 10 04 PM

confidant/schema/credentials.py
```diff
@@ -21,6 +21,7 @@ class CredentialResponse(object):
     permissions = attr.ib(default=dict)
     tags = attr.ib(default=list)
     last_rotation_date = attr.ib(default=None)
```
irhkang, line 23: not related to this PR but we could remove last_rotation_date if we aren't using it (which I don't think we should given that next_rotation_date has some complex logic)

If you remove it remove the line in the test too

confidant/public/modules/resources/views/credential-details.html
```diff
@@ -146,6 +146,10 @@
         <button type="button" ng-disabled="editableForm.$waiting" ng-click="addTag()" class="btn pull-right">Add Tag</button>
       </span>
     </div>
+    <div class="form-group" ng-show="credential.id">
+      <label for="credentialNextRotationDate">Credential Next Rotation Needed Date</label>
+      <span id="credentialNextRotationDate">{{ credential.next_rotation_date }} ({{ credential.daysTillRotation }} days)</span> <span id="credentialNextRotationDateTooltip" class="glyphicon glyphicon-info-sign" data-toggle="tooltip" title="This date is determined by the last rotation date and the last decryption date."></span>
```
irhkang, line 151: we ok with showing a negative number here for expired creds?

confidant/schema/credentials.py
ryan-lane: I'm ok with delivering it, even if it's not shown in the UI. We may decide to show it later, or folks may want to use it in other tooling

confidant/public/modules/resources/views/credential-details.html
ryan-lane: yep. I think it's good context that it should have been rotated a while ago.

PR #288, opened 2020-04-03
Minor bugfix to prevent duplicate tags

skiptomyliu:
```
"a" in ["a", "b", "c"]
false

0 in ["a", "b", "c"]
true

3 in ["a", "b", "c"]
false

"a" in {"a": "z"}
true

["a"].includes("a")
true
```

js is nuts to me.


ryan-lane:
Oh javascript, you're always giving me "wat" moments.

PR #289, opened 2020-04-03
Adding group attribute to credentials

laudiacay:
This PR adds a group field to credentials and users. If a user is not in the same group as a credential, they will not be able to interact with that credential at all. 

Not ready to merge currently- the branches diverged quite a bit in the past few months, and this feature is not completely integrated with credential archiving, tests, the UI, or code to check whether a credential has certain permissions. Currently working on this and will have it ready to go shortly.


confidant/authnz/__init__.py
```diff
@@ -90,6 +100,19 @@ def user_type_has_privilege(user_type, privilege):
...
+    # Fail open if:
+    # - the principal is not a human user
+    # - authz checks are disabled globally
+    # - OR there is no group specified
+    return (
+        not user_is_user_type('user')
+        or not app.config.get('USE_GROUPS')
+        or not groupname
+        or user_is_member(groupname)
+    )
```
ryan-lane, line 113: This should be implemented as an ACL check against actions of a resource: https://lyft.github.io/confidant/acls.html

confidant/authnz/userauth.py
```diff
@@ -113,13 +114,24 @@ def set_expiration(self):
...
             'last_name': last_name,
+            'groups': []
         }
 
+        if username is None:
+            username = email.strip().split('@')[0]
+
+        if app.config.get('USE_GROUPS'):
+            all_groups = grp.getgrall()
+            session['user']['groups'] = [g.gr_name for g in all_groups if email in g.gr_mem]
```
ryan-lane, line 133: I think it would be good if fetching groups was something modular, as it's possible that some folks will get the user's group from a SAML assertion, some will get it from a header, and others may want to use unix groups like this.

confidant/authnz/userauth.py
```diff
@@ -131,6 +143,13 @@ def current_first_name(self):
     def current_last_name(self):
         return self.current_user()['last_name']
 
+    def current_groups(self):
+        groups = self.current_user().get('groups')
+        if groups is None:
+            return []
+        else:
+            return groups
```
ryan-lane, line 151:
```suggestion
        groups = self.current_user().get('groups', [])
```

confidant/authnz/userauth.py
```diff
@@ -299,22 +319,27 @@ def assert_headers(self):
...
+        if self.username_header and self.username_header in request.headers:
+            username = request.headers[self.username_header]
+
+        # Set current user on every request to ensure that the user's group
+        # list is updated (TODO: caching?)
+        self.set_current_user(
+            request.headers[self.email_header],
+            first_name = first_name,
+            last_name = last_name,
+            username = username
```
ryan-lane, line 340: first_name and last_name here are changing their defaults from '' to None. Will this cause typing issues anywhere?

confidant/public/modules/resources/views/credential-details.html
```diff
@@ -38,6 +38,15 @@
...
       <span editable-checkbox="credential.enabled" id="credentialEnabled">{{ credential.enabled }}</span>
     </div>
+    <div class="form-group">
+      <label for="credentialGroup">Credential Group</label>
+      <span editable-select="credential.group"
+	    e-ng-options="x for x in groups( {{user.groups}} )"
+	    id="credentialGroupInput">
+	{{ credential.group || 'No group set' }}
+      </span>
+      <span ng-show="credential.id && !credential.authorized">YOU ARE NOT AUTHORIZED TO VIEW OR MODIFY THIS CREDENTIAL'S DETAILS</span>
```
ryan-lane, line 48: We return permission hints from individual resource level endpoints, for ACL checks; see: https://lyft.github.io/confidant/api.html#get--v1-credentials-(id)

confidant/routes/v1.py
```diff
@@ -45,9 +45,9 @@ def get_user_info():
     Get the email address of the currently logged-in user.
     '''
     try:
-        response = jsonify({'email': authnz.get_logged_in_user()})
+        response = jsonify({'email': authnz.get_logged_in_user(), 'groups': authnz.get_user_groups()})
     except authnz.UserUnknownError:
-        response = jsonify({'email': None})
+        response = jsonify({'email': None, 'groups': None})
```
ryan-lane, line 50: I'm confused about this diff, because `confidant/routes/v1.py` doesn't exist anymore. It's been split up into files specific to the types of endpoints.

confidant/routes/v1.py
```diff
@@ -283,6 +283,22 @@ def map_service_credentials(id):
...
+        # any credentials they have access to, even if the service has others
+        # assigned that they can't access
+        unauthorized_creds = []
+        for cred in _get_credentials(added + removed):
+            if not cred.get('authorized', False):
+                unauthorized_creds.append(cred['name'])
+        if len(unauthorized_creds) > 0:
+            msg = 'User not authorized to modify credentials: {0}'
+            ret = {'error': msg.format(", ".join(unauthorized_creds))}
+            return jsonify(ret), 403
```
ryan-lane, line 300: This should be handled via the new built-in ACL support.

confidant/routes/v1.py
```diff
@@ -377,21 +392,27 @@ def get_credential(id):
...
+    if authnz.user_is_authorized(cred.group):
+        _is_authorized = True
+        data_key = keymanager.decrypt_datakey(
+            cred.data_key,
+            encryption_context={'id': context}
+        )
+        cipher_version = cred.cipher_version
+        cipher = CipherManager(data_key, cipher_version)
+        _credential_pairs = cipher.decrypt(cred.credential_pairs)
+        _credential_pairs = json.loads(_credential_pairs)
```
ryan-lane, line 415: The newer version of the code handles this logic currently. It will only deliver decrypted credentials if the correct ACL permissions exist.

confidant/routes/v1.py
```diff
@@ -644,13 +675,20 @@ def _lowercase_credential_pairs(credential_pairs):
...
     data = request.get_json()
     if not data.get('documentation') and settings.get('ENFORCE_DOCUMENTATION'):
         return jsonify({'error': 'documentation is a required field'}), 400
     if not data.get('credential_pairs'):
         return jsonify({'error': 'credential_pairs is a required field'}), 400
     if not isinstance(data.get('metadata', {}), dict):
         return jsonify({'error': 'metadata must be a dict'}), 400
+    if app.config.get('USE_GROUPS'):
+        if data.get('group') and not authnz.user_is_member(data.get('group')):
+            return jsonify({'error': 'Must be a member of the destination group'}), 400
```
ryan-lane, line 691: This should be part of a custom ACL check.

confidant/routes/v1.py
```diff
@@ -779,6 +824,12 @@ def update_credential(id):
     cipher = CipherManager(data_key['plaintext'], version=2)
     credential_pairs = cipher.encrypt(update['credential_pairs'])
     update['metadata'] = data.get('metadata', _cred.metadata)
+    update['group'] = data.get('group')
+    # if using groups for access control, check that the user is a member
+    # of the new group. Setting the group to empty is always permitted
+    if app.config.get('USE_GROUPS'):
+        if update['group'] and not authnz.user_is_member(update['group']):
+            return jsonify({'error': 'Must be a member of the destination group'}), 400
```
ryan-lane, line 832: ACL check here as well.

confidant/public/modules/resources/views/credential-details.html
```diff
@@ -38,6 +38,15 @@
       <label for="credentialEnabled">Credential Enabled</label>
       <span editable-checkbox="credential.enabled" id="credentialEnabled">{{ credential.enabled }}</span>
     </div>
+    <div class="form-group">
```
ryan-lane, line 41: Since there's a server-side setting for whether groups are enabled or not, it would be good to only display this section if it's enabled. You can send the configuration option back to the client via the client config (https://github.com/lyft/confidant/blob/master/confidant/routes/identity.py#L127-L138). You can access the client config in the view and controller. It's already pulled into the scope; see this for example: https://github.com/lyft/confidant/blob/master/confidant/public/modules/resources/controllers/CredentialDetailsCtrl.js#L52-L53

laudiacay:
Going to redo this to work with the current version- branches diverged too much.

PR #291, opened 2020-04-17
mapping .:/srv/confidant seems like debugging leftover

f0rk:
alternatively, `make up` is incorrect and needs to run npm and grunt to
ensure the javascript/css is available


ryan-lane:
Ah, yeah. This probably shouldn't be here. Thanks for the PR!

PR #292, opened 2020-04-20
Remove legacy orca code for all DynamoDB tables

jfernandoz:
These tables definition already live in https://github.com/lyft/dsterraform/blob/master/orchestration/confidant/dynamodb.yaml. Most likely our refactorator change didn't remove it because it doesn't follow the most common naming to store salt states (ops/orca/states/).

### DIFF
--- a/salt/orchestration/confidant.sls
+++ b/salt/orchestration/confidant.sls
@@ -11,33 +11,6 @@
 {% endif %}
 {% endfor %}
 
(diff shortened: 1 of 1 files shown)


jfernandoz:
👀 @skiptomyliu (confidant oncall)

ryan-lane:
This is example code in an open source repo. You're making changes in the wrong spot

Issue #296, opened 2020-07-10
if HISTORY_PAGE_LIMIT isn't set, then the UI will fetch no history records


ryan-lane:
The UI sends in limit=0 to the backend, when it should be sending limit=null (or not setting the limit argument). Another approach is to have the backend unset limit if it's 0.

ryan-lane:
Fixed, and released in 6.3.0.

PR #297, opened 2020-07-10
Set history limit to None, if unset

ryan-lane:
HISTORY_PAGE_LIMIT uses int_env, which defaults to 0, rather than None. If the limit is set to 0, it'll be sent to the client as an int, which will request history with a limit of 0, rather than None. This change sets HISTORY_PAGE_LIMIT to None if unset, which correctly handles unlimited paging sizes.


ryan-lane:
Fixes https://github.com/lyft/confidant/issues/296

PR #332, opened 2021-12-24
Bug fix in logic that allows checking if a user is allowed to decrypt a credential

skiptomyliu:
When grabbing metadata of a credential, also check if the user has permissions to decrypt the credential.

This allows us to appropriately show or hide the lock.

### DIFF
--- a/confidant/routes/credentials.py
+++ b/confidant/routes/credentials.py
@@ -180,7 +180,11 @@ def get_credential(id):
 
     permissions = {
         'metadata': True,
-        'get': False,
+        'get': acl_module_check(
+            resource_type='credential',
+            action='get',
+            resource_id=id
+        ),
         'update': acl_module_check(
             resource_type='credential',
             action='update',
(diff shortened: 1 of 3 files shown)


confidant/routes/credentials.py
```diff
@@ -180,7 +180,11 @@ def get_credential(id):
 
     permissions = {
         'metadata': True,
-        'get': False,
```
meng-han, line 183: Question: what does True and False mean here?

meng-han (approved):

confidant/routes/credentials.py
skiptomyliu: True means that they're able to view certain UI components that allow the user to decrypt

eg:
https://github.com/lyft/confidant/blob/eeff4705cc29b6c8f873860378e33956af1fad24/confidant/public/modules/resources/views/credential-details.html#L54

Note this is only UI control.  The backend server ultimately allows / denies the decryption.

PR #334, opened 2021-12-29
Don't show decrypt icon if user does not have get permissions

skiptomyliu:
Users were being incorrectly shown decrypt icon on credentials where they had the ability to `create` credentials but didn't have `get` (decrypt) permissions.  

This change shows the option to decrypt credentials only if they have  decrypt permissions


achantavy (approved):

skiptomyliu:
💨

PR #343, opened 2022-10-11
Support generating signed JWTs

erickduran:
- Implements JWKManager which handles public and private keys
- Automatically imports key/cert pairs from CA_AUTHORITIES, if defined
- Adds route so authenticated services can retrieve signed JWTs
- Adds route to get public JWKS for environments that have set up keys

(keys in tests are self-signed and not used in any environment)


confidant/routes/jwks.py
```diff
@@ -0,0 +1,112 @@
...
+       GET /v1/jwks/public/staging
+
+    **Example response**:
+
+    .. sourcecode:: http
+
+       HTTP/1.1 200 OK
+       Content-Type: application/json
+
+       {
```
skiptomyliu, line 94: format should be

```
{"keys":
       [
        {"kty": "RSA",
         "kid": "123",
...
        },
{"kty": "RSA",
         "kid": "456",
...
        }
       ]
     }
```

See https://www.ietf.org/rfc/rfc7517.txt `A.1.  Example Public Keys`

confidant/routes/jwks.py
```diff
@@ -0,0 +1,112 @@
...
+
+    :resheader Content-Type: application/json
+    :statuscode 200: Success
+    :statuscode 400: JWTs are not supported for this user
+    """
+    user = authnz.get_logged_in_user()
+    if authnz.user_is_service(user):
+        payload = {
+            'service': user,
+            'namespace': user.replace('-iad', '')
```
skiptomyliu, line 46: we can't make the assumption that this will end in `-iad`

confidant/routes/jwks.py
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        if 'development' in user or 'unauthenticated' in user:
+            token = jwk_manager.get_jwt('localdev', payload)
```
skiptomyliu, line 50: need to remove this, as this is lyft specific.  we should pass this is in as parameter

confidant/services/jwkmanager.py
```diff
@@ -0,0 +1,97 @@
...
+            json.dumps(payload, sort_keys=True).encode('utf-8')
+        ).hexdigest()
+        if kid not in self._token_cache:
+            self._token_cache[kid] = {}
+
+        if payload_hash not in self._token_cache[kid]:
+            self._token_cache[kid][payload_hash] = jwt.encode(
+                payload=payload,
+                headers={'kid': kid},
+                key=key.export_to_pem(private_key=True, password=None),
```
Yueren-Wang, line 54: Shouldn't we allow customized password here during export. because on line 30-31, we check whether the passphrase is used when load from pem.

or we might remove passphrase check in 30-31 if we know we do not use password to wrap the key when export?

erickduran: On line 30-31 we use the passphrase to import into the JWKSet, that is where we're decrypting the key if provided as encrypted. On line 54 we're exporting it unencrypted so we can send it as input to the `jwt.encode()` function, I don't think there's an option to call it with an encrypted key, so we need to export it without a passphrase.

Also on 54 we're explicitly using `password=None` since the export function requires it to be done that way when you want no password when exporting.

Yueren-Wang: Got it. So you mean when we load from pem, it is possble that we need a password to decode, but when we export, we never encrypt the private key bytes.

the reason I am asking is in our test it looks like we are not testing set_key with a password, not sure whether it is a possible scenario tho

erickduran: Yes that's correct. We are not testing right now but it's a good point, I can add another scenario using a private key with password.

Yueren-Wang (dismissed):

confidant/settings.py
```diff
@@ -604,6 +605,15 @@ def str_env(var_name, default=''):
 # last saw a credential pair
 ENABLE_SAVE_LAST_DECRYPTION_TIME = bool_env('ENABLE_SAVE_LAST_DECRYPTION_TIME')
 
+# Add any certificate authorities
+if str_env('SERVICE_INSTANCE') == 'development':
```
skiptomyliu, line 609: `SERVICE_INSTANCE` is lyft specific.  I don't think we need this if clause?

erickduran: Good point, removing

confidant/services/jwkmanager.py
```diff
@@ -0,0 +1,109 @@
...
+            raise ValueError('This private key is not stored!')
+
+        now = datetime.now(tz=timezone.utc)
+        payload.update({
+            'iat': now,
+            'nbf': now,
+            'exp': now + timedelta(seconds=expiration_seconds),
+        })
+
+        payload_hash = sha1(
```
skiptomyliu, line 55: I think this is going to be expensive here, calculating sha hashes on each jwt fetch.  I'm thinking we can use the downstream service role name only as the "key".  Then the values includes the expiry and the JWT.  

Does the payload currently include the requesting service role name?

erickduran: Actually I think this part doesn't make a lot of sense any more (this was here before adding the expiry field). The purpose of this was to cache tokens so we don't generate on each call. Do we want to keep caching tokens for X amount of time? (e.g. send back the same token for its lifespan?)

skiptomyliu: I feel having a lever to either disable or enable it would be a good idea.  

It's unclear to me the CPU intensity to generate the tokens at moment (cryptographic functions usually have higher CPU requirements), but allowing us to easily enable/disable will allow us to experiment and enable us to keep CPU down (at expense of some memory).

erickduran: Added, the only downside I noticed is that caching this way will cause up to (# of replicas * # of workers) tokens at the same time for the same user, unless we support a distributed caching tool (since right now each worker has its own cache in memory)

skiptomyliu (approved):

PR #353, opened 2022-12-13
Add private key cache and remove unused code

erickduran:
Adding cache for private keys and removing unused code

### DIFF
--- a/confidant/services/jwkmanager.py
+++ b/confidant/services/jwkmanager.py
@@ -1,10 +1,7 @@
 import jwt
 
 from jwcrypto import jwk
-from typing import Any, Dict, Optional
-from hashlib import sha1
-from cryptography.x509 import load_pem_x509_certificate
-from cryptography.hazmat.primitives import serialization
+from typing import Dict, Optional
 from confidant.settings import CERTIFICATE_AUTHORITIES, \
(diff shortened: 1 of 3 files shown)


confidant/services/jwkmanager.py
```diff
@@ -38,10 +34,21 @@ def set_key(self, kid: str, private_key: str,
             self._keys.add(key)
         return kid
 
+    def _get_key(self, kid: str):
+        if kid not in self._pem_cache:
+            # setting either way to avoid further lookups when response is None
+            self._pem_cache[kid] = self._keys.get_key(kid)
+            if self._pem_cache[kid]:
```
skiptomyliu, line 41: should we remove the`if` statement?   This reads that `get_key(kid)` can return `None`.    If the pem cache is not set, we should just directly set it to the exported pem.

erickduran: if we remove the if this will throw an exception when trying to call `export_to_pem`. It's set up like this because we want `self._pem_cache[kid]` to return `None` if we have already tried to find that key to prevent keep looking for it on each call if it doesn't exist in `self._keys`

skiptomyliu: got it, so we're caching the result `None` here.

skiptomyliu (approved):

PR #354, opened 2022-12-16
Support multiple keys per environment

erickduran:
- Supports new format for CERTIFICATE_AUTHORITIES where `name` is environment (and multiple keys can be added):
  ```
  [
  {
    'key': redacted,
    'crt': redacted,
    'name': 'staging',
    'passphrase': redacted, 
    'kid': 'staging_slot_a'
  }, ...
  ]
  ```
- Signing key must be activated for each environment as `ACTIVE_SIGNING_KEYS={"staging": "some_kid", "production": "some_kid"}`, otherwise endpoint will return 400. This will be the key used to sign tokens.
- Public JWKS will expose all keys per environment.

### DIFF
--- a/confidant/services/jwkmanager.py
+++ b/confidant/services/jwkmanager.py
@@ -63,9 +63,11 @@ class JWKManager:
 
         if kid not in self._pem_cache[environment]:
             # setting either way to avoid further lookups when response is None
-            self._pem_cache[environment][kid] = self._keys[environment].get_key(kid)
+            self._pem_cache[environment][kid] = \
+                self._keys[environment].get_key(kid)
             if self._pem_cache[environment][kid]:
-                self._pem_cache[environment][kid] = self._pem_cache[environment][kid].export_to_pem(
+                self._pem_cache[environment][kid] = \
+                    self._pem_cache[environment][kid].export_to_pem(
                     private_key=True,
                     password=None
                 )
@@ -116,7 +118,6 @@ class JWKManager:
         }
         stats.incr('get_jwt.create')
         return token
-
 
     def get_active_key(self, environment: str) -> jwk.JWK:
         if environment in ACTIVE_SIGNING_KEYS and environment in self._keys:
--- a/confidant/services/jwkmanager.py
+++ b/confidant/services/jwkmanager.py
@@ -121,7 +121,7 @@ class JWKManager:
 
     def get_active_key(self, environment: str) -> jwk.JWK:
         if environment in ACTIVE_SIGNING_KEYS and environment in self._keys:
-            return self._keys[environment]._get_key(
+            return self._keys[environment].get_key(
                 ACTIVE_SIGNING_KEYS[environment],
                 environment
             )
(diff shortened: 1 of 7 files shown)


confidant/services/jwkmanager.py
```diff
@@ -1,65 +1,87 @@
...
-                             passphrase=ca['passphrase'])
-
-    def set_key(self, kid: str, private_key: str,
+            for environment in CERTIFICATE_AUTHORITIES:
+                for ca in CERTIFICATE_AUTHORITIES[environment]:
+                    if validator.validate(ca):
+                        self.set_key(environment, ca['kid'], ca['key'],
+                                     passphrase=ca['passphrase'])
+                    else:
+                        logger.error(f'Invalid entry in {environment} '
```
skiptomyliu, line 44: let's add the kid also.

confidant/settings.py
```diff
@@ -636,3 +636,4 @@ def get(name, default=None):
 # Module that will perform an external ACL check on API endpoints
 ACL_MODULE = str_env('ACL_MODULE', 'confidant.authnz.rbac:default_acl')
 DEFAULT_JWT_EXPIRATION_SECONDS = int_env('DEFAULT_JWT_EXPIRATION_SECONDS', 3600)
+ACTIVE_SIGNING_KEYS = json.loads(str_env('ACTIVE_SIGNING_KEYS', '{}'))
```
skiptomyliu, line 639: can we add a comment above on format expected here?

confidant/services/jwkmanager.py
```diff
@@ -115,14 +137,24 @@ def get_payload(self, certificate: str, token: str,
...
+            return self._keys[environment].get_key(
+                ACTIVE_SIGNING_KEYS[environment]
+            )
+
+    def get_jwks(self, environment: str, algorithm: str = 'RS256') \
+            -> List[Dict[str, str]]:
+        keys = self._keys.get(environment)
+        if keys:
+            stats.incr(f'get_jwks.{environment}.hit')
+            return [{
```
skiptomyliu, line 151: I think we want it in this format instead:


```
{
  "keys": [
    {
      "kty": "RSA",
      "e": "...",
      "n": "...",
      "alg": "RS256",
      "kid": "staging_1"
    },
{
      "kty": "RSA",
      "e": "...",
      "n": "...",
      "alg": "RS256",
      "kid": "staging_2"
    }
  ]
}
```

erickduran: Yeah this is the final format (all fields are exported by `export_public()` except `alg`). The route itself wraps the keys in the `keys` object per the schema:
https://github.com/lyft/confidant/blob/master/confidant/schema/jwks.py#L23

skiptomyliu (approved):

confidant/settings.py
```diff
@@ -606,6 +606,17 @@ def str_env(var_name, default=''):
 ENABLE_SAVE_LAST_DECRYPTION_TIME = bool_env('ENABLE_SAVE_LAST_DECRYPTION_TIME')
 
 # Add any certificate authorities
+# Should be in encrypted settings following this format (where name is
+# the name of the environment) and key ids must be unique:
+# [
```
skiptomyliu, line 611: @erickduran just caught this now, I don't think this is the correct format anymore.

PR #357, opened 2023-01-03
Remove excessive permission checks for get service

meng-han:
This excessive permission check may introduce unnecessary warnings.


confidant/routes/services.py
```diff
@@ -209,16 +209,18 @@ def get_service(id):
...
-        action='get',
-        resource_id=id,
-    )
+    if action == 'metadata':
+        permissions['metadata'] = acl_module_check(
+            resource_type='service',
+            action='metadata',
+            resource_id=id,
+        )
+    elif action == 'get':
```
ruwaifaa, line 218:
```suggestion
    else:
```

meng-han: there's a 'update' in the permission map, so i keep it more strict

ruwaifaa (approved):

confidant/routes/services.py
meng-han:
```
  permissions = {
        'metadata': False,
        'get': False,
        'update': False,
    }
```

ruwaifaa: action can't take any other values besides "get" and "metadata" tho. np. just a nit tho

PR #358, opened 2023-01-26
Adding pagination for credentials

erickduran:
Adding pagination for credentials, keeping `limit` and `page` as optional so it's backward compatible (not specifying will keep previous behavior).

### DIFF
--- a/confidant/routes/credentials.py
+++ b/confidant/routes/credentials.py
@@ -123,7 +123,8 @@ def get_credential_list():
     else:
         credentials_response = CredentialsResponse.from_credentials([
             credential
-            for credential in Credential.data_type_date_index.query('credential')
+            for credential in
+            Credential.data_type_date_index.query('credential')
         ])
     return credentials_response_schema.dumps(credentials_response)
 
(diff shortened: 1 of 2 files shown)


confidant/routes/credentials.py
```diff
@@ -93,10 +93,39 @@ def get_credential_list():
...
+            'credential',
+            scan_index_forward=False,
+            limit=limit,
+            last_evaluated_key=page,
+        )
+        credentials_response = CredentialsResponse.from_credentials(
+            [credential for credential in results],
+            next_page=results.last_evaluated_key,
+        )
+    else:
```
skiptomyliu, line 123: do we want this to be a `if not page and not limit` ?  Otherwise this will override the `if page` branch if no `limit` is supplied.

erickduran: That's the goal, we don't want to paginate if no `limit` is given, so we ignore the `page` and just return everything. I can add an error response if just `page` is given and not a `limit`

skiptomyliu (approved):

PR #359, opened 2023-01-27
Adding pagination for services

erickduran:
Adding pagination for credentials, keeping `limit` and `page` as optional so it's backward compatible (not specifying will keep previous behavior).

### DIFF
--- a/confidant/routes/services.py
+++ b/confidant/routes/services.py
@@ -91,6 +91,7 @@ def get_service_list():
 
     :query string next_page: If paged results were returned in a call, this
                              query string can be used to fetch the next page.
+    :query int limit: Limit of items per each page (required for pagination)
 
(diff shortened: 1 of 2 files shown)


confidant/routes/services.py
```diff
@@ -128,9 +129,40 @@ def get_service_list():
...
+            'service',
+            scan_index_forward=False,
+            limit=limit,
+            last_evaluated_key=page,
+        )
+        services_response = ServicesResponse.from_services(
+            [result for result in results],
+            next_page=results.last_evaluated_key
+        )
+    else:
```
skiptomyliu, line 162: same feedback as https://github.com/lyft/confidant/pull/358/files#r1091107505

skiptomyliu (approved):

PR #360, opened 2023-01-27
Adding pagination for service credentials

erickduran:
Implement pagination (in a separate route) for service credentials.

Using a separate route to simplify logic and responses.


skiptomyliu:
@erickduran , i was under impression that we're going to paginate the listing of services and credentials.  This seems to paginate creds for services though?

PR #361, opened 2023-02-02
Adding JWT user mapping

erickduran:
- Adding dynamic import via `ACL_MODULE` to check for appropriate access
- Response JWT will replace the `user` with the provided `id` if the mapping allows it, otherwise will return 403
- Response JWT will include `parent` (the original authenticated username) moving forward

### DIFF
--- a/confidant/routes/jwks.py
+++ b/confidant/routes/jwks.py
@@ -42,12 +42,17 @@ def get_token():
     :statuscode 200: Success
     :statuscode 400: JWTs are not supported for this user
     """
-    user = authnz.get_logged_in_user()
+    parent = authnz.get_logged_in_user()
+    user = parent
     environment = request.args.get('environment', type=str)
 
     if not environment:
         return jsonify({'error': 'Please specify an environment'}), 400
 
+    payload = {
+        'is_service': authnz.user_is_service(user),
+        'parent': parent,
+    }
     target_user = request.args.get('target_user', type=str)
     if target_user and JWT_MAPPING_MODULE:
         if jwt_mapping_check(user, target_user):
@@ -58,10 +63,8 @@ def get_token():
             })
             return response, 403
 
-    payload = {
-        'user': user,
-        'is_service': authnz.user_is_service(user),
-    }
+    payload['user'] = user
+
     try:
         token = jwk_manager.get_jwt(environment, payload)
     except ValueError:
--- a/confidant/routes/jwks.py
+++ b/confidant/routes/jwks.py
@@ -6,18 +6,18 @@ from confidant import authnz
 from confidant.services.jwkmanager import jwk_manager
 from confidant.schema.jwks import jwt_response_schema, JWTResponse, \
     jwks_list_response_schema, JWKSListResponse
-from confidant.settings import JWT_MAPPING_MODULE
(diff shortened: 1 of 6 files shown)


confidant/routes/jwks.py
```diff
@@ -6,11 +6,14 @@
...
 from confidant.schema.jwks import jwt_response_schema, JWTResponse, \
     jwks_list_response_schema, JWKSListResponse
-
+from confidant.settings import JWT_MAPPING_MODULE
+from confidant.utils import misc
 
 logger = logging.getLogger(__name__)
 blueprint = blueprints.Blueprint('jwks', __name__)
 
+jwt_mapping_check = misc.load_module(JWT_MAPPING_MODULE)
```
skiptomyliu, line 15: let's conform to use `acl_module_check = misc.load_module(settings.ACL_MODULE)` instead.  This centralizes the RBAC logic.

confidant/routes/jwks.py
```diff
@@ -39,16 +42,29 @@ def get_token():
...
     if not environment:
         return jsonify({'error': 'Please specify an environment'}), 400
 
     payload = {
-        'user': user,
         'is_service': authnz.user_is_service(user),
+        'parent': parent,
     }
+    target_user = request.args.get('target_user', type=str)
+    if target_user and JWT_MAPPING_MODULE:
```
skiptomyliu, line 57: let's change this to

```bash
if not acl_module_check(resource_type='jwt', action='get', resource_id=id):
        msg = f"{authnz.get_logged_in_user()} does not have access to get JWT {id}"
        error_msg = {'error': msg}
        return jsonify(error_msg), 403
```

confidant/routes/jwks.py
```diff
@@ -6,11 +6,14 @@
...
+from confidant.utils import misc
 
 logger = logging.getLogger(__name__)
 blueprint = blueprints.Blueprint('jwks', __name__)
 
+jwt_mapping_check = misc.load_module(JWT_MAPPING_MODULE)
+
 
 @blueprint.route('/v1/jwks/token', methods=['GET'])
 @authnz.require_auth
```
skiptomyliu, line 19: we need to change this to

```python
def get_token(id):
```

the `id` will the be the token (resource_id) we're attempting to request.

confidant/routes/jwks.py
```diff
@@ -39,16 +42,23 @@ def get_token():
     :statuscode 200: Success
     :statuscode 400: JWTs are not supported for this user
     """
-    user = authnz.get_logged_in_user()
+    parent = authnz.get_logged_in_user()
```
skiptomyliu, line 45: should we change this to `host`?  trying to think of a better name 🤷

skiptomyliu: eh, should be fine for now.

skiptomyliu (dismissed):

skiptomyliu (approved):

confidant/routes/jwks.py
Yueren-Wang: so this id will be the envoy_service name?

skiptomyliu: @Yueren-Wang yes

PR #362, opened 2023-02-03
Add kwargs to JWT fetch

skiptomyliu:
We need two parameters in the JWT RBAC check: the logged in user AND the requested JWT id.

Minor variable naming changes.

### DIFF
--- a/confidant/routes/jwks.py
+++ b/confidant/routes/jwks.py
@@ -37,7 +37,7 @@ def get_token(id):
        {
          "is_service": true,
          "requester": "pyexample2",
-         "token": "ey..."
+         "user": "ey..."
        }
 
     :resheader Content-Type: application/json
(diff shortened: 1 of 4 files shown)


confidant/routes/jwks.py
```diff
@@ -35,27 +35,36 @@ def get_token(id):
        Content-Type: application/json
 
        {
-         "token": "ey..."
+         "is_service": true,
+         "requester": "pyexample2",
+         "user": "ey..."
        }
```
erickduran, line 41: these are the contents within the JWT payload, but we're still responding as 
```
"token": "ey..."
```

erickduran (approved):

confidant/routes/jwks.py
skiptomyliu: thanks.  corrected

erickduran (approved):

PR #366, opened 2023-02-04
Remove duplicate markupsafe from docs/requirements.txt

bwitt:
`MarkupSafe` is already pulled in from `../requirements.txt` so no need to explicitly include in `docs/requirements.txt`


skiptomyliu (approved):
thanks!

PR #371, opened 2023-02-08
Implement archive route

erickduran:
- Moving archive functions to service
- Implementing route for FE

### DIFF
--- a/confidant/routes/credentials.py
+++ b/confidant/routes/credentials.py
@@ -1135,28 +1135,15 @@ def generate_value():
 def archive_credential(id):
     '''
     Archive the provided credential.
-
-    .. :quickref: Credential; Update the provided credential using the data
-                  provided in the post body.
-
     **Example request**:
 
     .. sourcecode:: http
 
-       PUT /v1/credentials/abcd12345bf4f1cafe8e722d3860404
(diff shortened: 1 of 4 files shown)


confidant/routes/credentials.py
```diff
@@ -1126,3 +1126,69 @@ def generate_value():
...
+    try:
+        _cred = Credential.get(id)
+    except DoesNotExist:
+        return jsonify({'error': 'Credential not found.'}), 404
+    if _cred.data_type != 'credential':
+        msg = 'id provided is not a credential.'
+        return jsonify({'error': msg}), 400
+
+    data = request.get_json()
+    force = data.get('force') if data else False
```
skiptomyliu, line 1183:
```suggestion
    force = data.get('force', False) if data else False
```

skiptomyliu: dbl check me on this (i haven't ran any code), but we basically wanna fail safe and force explicitly setting the boolean to True if we want to archive.

skiptomyliu (approved):

skiptomyliu:
lgtm, had minor suggestion, i'll approve again once we address

skiptomyliu (approved):

PR #372, opened 2023-02-11
Adding new credentials to ACL module check for services

erickduran:


confidant/routes/services.py
```diff
@@ -579,12 +579,16 @@ def map_service_credentials(id):
     credentials = data.get('credentials', [])
     blind_credentials = data.get('blind_credentials', [])
     combined_credentials = credentials + blind_credentials
+    new_credentials = set(combined_credentials).difference(
+        _service.credentials + _service.blind_credentials
```
skiptomyliu, line 583: Does `_service.credentials` return a list of ids?  Making sure it does not return an object / dictionary of some sort.

erickduran: good catch, it's actually a set so just change to `union()`

skiptomyliu (approved):

PR #379, opened 2023-03-27
Add stats for signing key used

skiptomyliu:
Also minor refactor to prepend all jwt ca related stats with `jwt`.

### DIFF
--- a/confidant/services/jwkmanager.py
+++ b/confidant/services/jwkmanager.py
@@ -137,7 +137,7 @@ class JWKManager:
         # The active signing key used to sign JWTs
         if environment in JWT_ACTIVE_SIGNING_KEYS and environment in self._keys:
             kid = JWT_ACTIVE_SIGNING_KEYS[environment]
-            stats.incr(f'jwt.get_active_key.{kid}')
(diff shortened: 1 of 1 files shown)


confidant/services/jwkmanager.py
```diff
@@ -125,28 +130,28 @@ def get_jwt(self, environment: str, payload: dict,
...
 
     def get_active_key(self, environment: str) -> Tuple[str, Optional[jwk.JWK]]:
+        # The active signing key used to sign JWTs
         if environment in JWT_ACTIVE_SIGNING_KEYS and environment in self._keys:
-            return JWT_ACTIVE_SIGNING_KEYS[environment], self._get_key(
-                JWT_ACTIVE_SIGNING_KEYS[environment],
-                environment
-            )
+            kid = JWT_ACTIVE_SIGNING_KEYS[environment]
+            stats.incr(f'jwt.get_active_key.{environment}.{kid}')
```
skiptomyliu, line 140: this is the new stats that we want.  it will allow us to see certificate key id used for signing

Yueren-Wang (approved):

PR #393, opened 2023-08-04
Fix JWT cache

skiptomyliu:
- Refactor JWT cache code.  Mostly moving into a separate class that will strictly handle cache logic.  Add an abstract JwtCache class so it's easy to extend to an eventual remote redis call.
- Introduce `cachetools`.
- Add `JWT_CACHING_TTL_SECONDS` which allows us to specify when tokens should be evicted.  Before, it was possible that Confidant would provision expired JWTs from its custom cache.


confidant/services/jwkmanager.py
```diff
@@ -22,15 +25,102 @@
...
+            stats.incr('jwt.get_jwt.cache.miss')
+        return jwt_str
+
+    def set_jwt(self, kid: str, requester: str, user: str, expiry: str,
+                jwt: str) -> None:
+        if kid not in self._token_cache:
+            self._token_cache[kid] = {}
+
+        if requester not in self._token_cache[kid]:
+            self._token_cache[kid][requester] = {}
```
ruwaifaa, line 98: We can shove the entire thing as the dictionary key. it will make it the code simpler
`key = f'jwt:{kid}:{requester}:{user}'`

confidant/services/jwkmanager.py
```diff
@@ -22,15 +25,102 @@
 CA_SCHEMA = {
     'crt': {'type': 'string', 'required': True},
     'key': {'type': 'string', 'required': True},
-    'passphrase': {'type': 'string', 'required': True},
+    'passphrase': {'type': 'string', 'required': True, 'nullable': True},
```
ruwaifaa, line 28: What do we use this for?

ruwaifaa:
We should keep the size bounded of the cache. Should we use some library like cachetools for that?

confidant/services/jwkmanager.py
skiptomyliu: good idea, i'll update

confidant/services/jwkmanager.py
skiptomyliu: https://github.com/lyft/confidant/blob/96c5f5bd907efe03f2885c2e8eb42c28bf0f1f4f/confidant/services/jwkmanager.py#L43

It fails if `passphrase` is None, which it currently is in our unit tests https://github.com/lyft/confidant/blob/96c5f5bd907efe03f2885c2e8eb42c28bf0f1f4f/tests/conftest.py#L159

It's failing after I updated our unit tests to actually hit this path.

skiptomyliu:
> We should keep the cache size bounded. Should we use some library like cachetools for that? This is one such library that implements cache with TTL https://github.com/tkem/cachetools

added `cachetools`

confidant/services/jwkmanager.py
```diff
@@ -8,29 +8,90 @@
...
+    @abstractmethod
+    def get_jwt(self, kid: str, requester: str, user: str) -> str:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def set_jwt(self, kid: str, requester: str, user: str, jwt: str) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def cache_key(self, kid: str, requester: str, user: str) -> str:
```
ruwaifaa, line 56: nit: Probably not needed as in abstract base class as it can be an internal function of derived classes

ruwaifaa (approved):

confidant/services/jwkmanager.py
skiptomyliu: removed

ruwaifaa (approved):

PR #424, opened 2024-04-08
Profiling credentials

alejandroroiz:
Add a timer stat to update and create credentials


confidant/routes/credentials.py
```diff
@@ -592,92 +592,95 @@ def create_credential():
                      correct format, or a required field was not provided.
     :statuscode 403: Client does not have access to create credentials.
     '''
-    if not acl_module_check(resource_type='credential', action='create'):
-        msg = "{} does not have access to create credentials".format(
-            authnz.get_logged_in_user()
+    with stats.timer('create_credential'):
+        if not acl_module_check(resource_type='credential', action='create'):
+            msg = "{} does not have access to create credentials".format(
```
meng-han, line 597: nit: it would be good to start using f-strings in confidant. no need to change it for now but at least code that we touch in the future or add.

meng-han (approved):

confidant/routes/credentials.py
alejandroroiz: good point, will start focusing on this

PR #425, opened 2024-04-17
Further Sanitize User Input

alejandroroiz:
In this PR we fix the following issues:

- Use f-strings in the create/update credential routes 
- Escape metadata in credential creation
- Check if the credential name was updated
  - If it was then verify it does not conflict with an existing credential name
  - Also sanitize the new credential name
- Check if new metadata was added and sanitize it
- Check if new credential pairs were added and sanitize them
- Escape documentation field in update & create


anshumanbh:
Also, cc: @diegonavarro-lyft @0xp4blo

anshumanbh (approved):

confidant/routes/credentials.py
```diff
@@ -816,6 +820,34 @@ def update_credential(id):
...
+        # existing credential and to ensure we don't escape the name if it
+        # hasn't changed
+        if data.get('name') != _cred.name:
+            data['name'] = escape(data.get('name'))
+            for cred in Credential.data_type_date_index.query(
+                    'credential',
+                    filter_condition=Credential.name == data['name']):
+                # Conflict, the name already exists
+                msg = f'Name already exists. See id: {cred.id}'
+                return jsonify({'error': msg, 'reference': cred.id}), 409
```
meng-han, line 833: what about the latency here? i remember you mentioned a high latency for check dup

meng-han (approved):

confidant/routes/credentials.py
alejandroroiz: the whole dup check adds more latency since we need to decrypt credential pairs to verify if the cred is a full duplicate. this check is only for dup name, which is not as expensive

Yueren-Wang (approved):

PR #431, opened 2024-08-20
add timer to key GET paths

meng-han:
Was wondering what caused frontend to be so slow? did an analysis on UI and ranked captured network calls by timer. The slow and popular APIs are get and list services/credentials. Emitting stats to understand how slow it is and follow up on action items

### DIFF
--- a/confidant/routes/credentials.py
+++ b/confidant/routes/credentials.py
@@ -86,48 +86,49 @@ def get_credential_list():
     :statuscode 200: Success
     :statuscode 403: Client does not have permissions to list credentials.
     """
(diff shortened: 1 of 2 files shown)


confidant/routes/credentials.py
```diff
@@ -86,48 +86,49 @@ def get_credential_list():
     :statuscode 200: Success
     :statuscode 403: Client does not have permissions to list credentials.
     """
```
meng-han, line 88: it's interesting how git detects the diff: the only changes made is add `with stats.timer(` and added indent to following lines

alejandroroiz (approved):

PR #434, opened 2024-08-27
add logging abstraction

alejandroroiz:
no longer needed

----

Added logging module abstraction. The logging module to be used by setting the `LOGGING_MODULE` environment variable.

Also shifted some code around to avoid circular dependencies.

## Testing

- [x] local dev works
- [x] logs still work as expected
- [x] all actions can be executed


alejandroroiz:
no longer needed

PR #435, opened 2024-08-30
Change test saml port in docker compose

whu-lyft:
Fix the local confidant login issue


alejandroroiz:
this is not needed, please don't try to change configurations for everybody to fix an issue that is only affecting your setup.

PR #436, opened 2024-09-11
Prevent XSS from API call

whu-lyft:
### Description: 
Prevent XSS

### Test
test in confidant-staging


confidant/utils/misc.py
```diff
@@ -54,3 +55,24 @@ def utcnow():
     """
     now = datetime.utcnow()
     return now.replace(tzinfo=pytz.utc)
+
+
+def prevent_xss(pre_xss_response):
```
meng-han, line 60: recommend to change this as a decorator, just an example like this

meng-han:
```
def prevent_xss_decorator(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        # Call the original function to get the response
        pre_xss_response = func(*args, **kwargs)
        
        # Apply XSS prevention
        response = make_response(pre_xss_response)
        response.headers['Content-Type'] = 'application/json'
        response.headers['X-XSS-Protection'] = '1; mode=block'
        response.headers['X-Content-Type-Options'] = 'nosniff'
        response.headers['Referrer-Policy'] = 'no-referrer'
        
        return response
    return wrapper
```

meng-han: then use it in the routes that can be applied `@prevent_xss_decorator` above the route declaration

meng-han (commented):
i'll leave another comment in private channels

confidant/utils/misc.py
```diff
@@ -54,3 +56,32 @@ def utcnow():
...
+     5. Limit Referrer Information
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        # Call the original function to get the response
+        pre_xss_response = func(*args, **kwargs)
+
+        # If the response is not a dict, return it as-is
+        if not isinstance(pre_xss_response, dict):
+            return pre_xss_response
```
meng-han, line 77: let's try if we can remove this block. Flask's make_response should be able to handle the case of '{}, status_code', see https://flask.palletsprojects.com/en/3.0.x/api/#flask.make_response

meng-han (approved):

PR #442, opened 2024-10-31
Send select logs to Panther

masoudv-lyft:
Sends a few important logs types to Panther for detection.

### DIFF
--- a/confidant/authnz/userauth.py
+++ b/confidant/authnz/userauth.py
@@ -23,6 +23,7 @@ from confidant import settings
 from confidant.lib import cryptolib
 from confidant.utils.misc import dict_deep_update
 from confidant.authnz import errors
+from confidant.services.panther import panther_client
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


masoudv-lyft:
@alejandroroiz The 0.1.4 release of the library is available: https://artifactory.lyft.net/ui/packages/pypi:%2F%2Flyft-lumos-common/0.1.4?name=lumos&type=packages

But the checks are failing, is there a right way I should be injecting this dependency? Seems to be doing fine here: https://github.com/lyft/confidant-private/pull/1613

confidant/authnz/userauth.py
```diff
@@ -586,7 +587,11 @@ def consume_saml_assertion(self):
 
         attributes = auth.get_attributes()
         logger.info('SAML attributes: {!r}'.format(attributes))
-
+        panther_client.send_event({
```
masoudv-lyft, line 590: Fail open calls to Panther:

https://github.com/lyft/python-lyft-lumos-common/blob/main/lyft_lumos_common/services/panther.py#L42

